PR #68, opened 2019-04-25
General Improvements

nhooyr:


nhooyr:
Decided against the Conn API for helpers because I don't like how the wrapping makes fields ambiguous. E.g. whether JSONConn should embed Conn or not and if not, should we provide helpers for Close and Subprotocol that call to the underlying Conn? A plain function is just so much simpler.

PR #81, opened 2019-05-17
Refactoring

### LINKED ISSUE #78: Improve Close UX
https://www.reddit.com/r/golang/comments/bi2tan/nhooyrwebsocket_a_minimal_and_idiomatic_websocket/em1jn6d/?context=3

The close only happens for the first call.

Also, instead of returning an error when the close code/reason are invalid, we should just print to stderr like net/http as no one will ever check Close's error, in normal usage anyway.

nhooyr:
Closes #78


nhooyr:
This PR adds concurrent reads and massively simplifies the internals of the libraries 👏

nhooyr:
Figured out a clever way to merge readLoop and timeoutLoop into a single goroutine!

nhooyr:
Nvm I did not, I thought io.Reader was non blocking but it isn't. It blocks for at least a single read usually. So if we combine the two, then there is no way to get readLoop to wake up.

nhooyr:
@coadler Definitely let me know how you feel about the additions to the API.

coadler:
:+1: will take a look tomorrow morning

coadler:
I'll test these out on my gateway later this weekend and let you know anything else I find

PR #99, opened 2019-06-23
Add CloseRead and closeError test

nhooyr:
@ssmith so I added a test for the issue you reported to me.

It passes. Does your code look any different?

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -74,11 +74,7 @@ func Example_writeOnly() {
 		ctx, cancel := context.WithTimeout(r.Context(), time.Minute*10)
 		defer cancel()
 
-		go func() {
-			defer cancel()
-			c.Reader(ctx)
-			c.Close(websocket.StatusPolicyViolation, "server doesn't accept data messages")
-		}()
+		ctx = c.CloseRead(ctx)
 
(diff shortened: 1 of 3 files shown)


nhooyr:
Whoops meant to notify you. @scsmithr

scsmithr:
So it turns out I was just misplacing a `defer cancel()`.

If I change your "closeError" test to the following, I get the EOF.
```
server: func(w http.ResponseWriter, r *http.Request) error {
				c, err := websocket.Accept(w, r, websocket.AcceptOptions{})
				if err != nil {
					return err
				}
				defer c.Close(websocket.StatusInternalError, "")

				ctx, cancel := context.WithCancel(r.Context())
				defer cancel()

				go func() {
					defer cancel()
					c.Reader(ctx)
					c.Close(websocket.StatusPolicyViolation, "")
				}()

				err = wsjson.Write(ctx, c, "hello")
				if err != nil {
					return err
				}

				return nil
			},
```

Removing the extra `defer cancel()` fixes it.

nhooyr:
Makes sense, you're closing the connection early with the defer before the close frame.

PR #112, opened 2019-08-06
Problem: NetConn wrapper kept canceling context

### LINKED ISSUE #111: Weird behaviour when channelling go-capnproto2 over websocket
My setup is a little convoluted. I use capnproto rpc over websockets (to simplify deployment) between my programs. However, the websocket server keeps failing to read a header somehow and crashes.

```
rpc: writing return: failed to write msg: websocket closed: failed to read header: EOF
```

and on the Dialer side:

```
rpc: writing finish: failed to write msg: context canceled
```

Sometimes both ends cancel their contexts at the same time. I'm not sure why.

Vaelatern:
Solution: Accept the zero time as a special case when setting
deadlines. Somehow, this function gets called without a time and thus
would previously Reset the timer to no time at all, thus expediting the
call to cancel the context.

Would close #111

### DIFF
--- a/netconn.go
+++ b/netconn.go
@@ -134,11 +134,23 @@ func (c *netConn) SetDeadline(t time.Time) error {
 }
 
 func (c *netConn) SetWriteDeadline(t time.Time) error {
-	c.writeTimer.Reset(t.Sub(time.Now()))
+	if t.IsZero() {
+		if !c.writeTimer.Stop() {
+			<-c.writeTimer.C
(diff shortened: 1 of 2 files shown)


nhooyr:
Thank you so much @Vaelatern 

I've merged this as https://github.com/nhooyr/websocket/commit/fb30712af122f689784e91a2723e2cd511ccd2d3 and will tag a new release.

PR #133, opened 2019-09-02
Add a CloseError example

### LINKED ISSUE #128: Improve CloseError API
I think the current API is a little obnoxious since it forces you to use `xerrors.As` to get the close error code which is very annoying.

Also people forget to use `xerrors.As`, see https://github.com/sumorf/deribit-api/blob/6b581052e02d0808bed0bdc7a04f3dc4582e70d3/stream.go#L30

@sumorf

I don't like the ceremony involved with `xerrors.As` so I'd like to provide some sort of convenience API to get the error code.

nhooyr:
Closes #128


nhooyr:
Appreciated the review @andersfylling

PR #135, opened 2019-09-03
Remove extra dependencies

### LINKED ISSUE #134: Remove dependency on golang.org/x
See https://github.com/nhooyr/websocket/issues/128#issuecomment-527643088

Would be nice to be truly zero dep.

Right now we depend on httpguts and xerrors.

xerrors can be easily removed now that Go 1.13 is out. Only problem is no frame information anymore which is unfortunate but welp

For httpguts I think it can be easily removed too.

nhooyr:
Closes #134

httpguts is an almost direct port of https://github.com/golang/net/blob/master/http/httpguts/httplex.go#L106

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   fmt:
     docker:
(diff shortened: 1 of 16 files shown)


nhooyr:
Waiting on Go 1.13 to get into home-brew.

PR #140, opened 2019-09-20
Switch to GitHub Actions for CI

### LINKED ISSUE #138: Try out GitHub actions again
Much nicer UI than CircleCi and both require custom images so no real difference when it comes to caching for me.

nhooyr:
Closes #138


nhooyr:
Half as fast but twice as beautiful so it evens out 😛

PR #142, opened 2019-09-21
Implement core API for WASM

### LINKED ISSUE #121: Compile for WebAssembly
@albrow I saw you in issue https://github.com/gorilla/websocket/issues/432#issuecomment-518355272

If you'd like to contribute the necessary improvements to this library, I'd be more than happy to help, review and merge your changes.

nhooyr:
Closes #121


nhooyr:
@albrow is it possible you could review this?

nhooyr:
@coadler pls review if you can

internal/wsecho/wsecho.go
```diff
@@ -0,0 +1,73 @@
...
+// The read limit is set to 1 << 40.
+func Loop(ctx context.Context, c *websocket.Conn) {
+	defer c.Close(websocket.StatusInternalError, "")
+
+	c.SetReadLimit(1 << 40)
+
+	ctx, cancel := context.WithTimeout(ctx, time.Minute)
+	defer cancel()
+
+	b := make([]byte, 32768)
```
coadler: Since you used `1 << 40` above `32 << 10` would be consistent here

albrow:
Sure. Should be able to review tomorrow.

internal/wsecho/wsecho.go
nhooyr: Also just realized 1 << 40 is a terabyte, not a gigabyte lol.

ci/wasm.sh
```diff
@@ -5,7 +5,23 @@ cd "$(dirname "${0}")"
...
+if [[ -z $WS_ECHO_SERVER_URL ]]; then
+  echo "./internal/wsecho/cmd/wsecho failed to start in 10s"
+  exit 1
+fi
+
+go install github.com/agnivade/wasmbrowsertest
+GOOS=js GOARCH=wasm go test -exec=wasmbrowsertest ./... -args "$WS_ECHO_SERVER_URL"
+
+kill %1
+wait -n || true
```
nhooyr: Will improve this test to assert proper closure of the connection.

nhooyr:
Going to merge for now since I want to restructure the codebase but please review if you can @albrow. I'll fix anything you point out afterwards.

internal/wsjs/wsjs.go
```diff
@@ -57,29 +59,33 @@ type WebSocket struct {
...
 
-func (c *WebSocket) BufferedAmount() uint32 {
-	return uint32(c.v.Get("bufferedAmount").Int())
-}
-
-func (c *WebSocket) addEventListener(eventType string, fn func(e js.Value)) {
-	c.v.Call("addEventListener", eventType, js.FuncOf(func(this js.Value, args []js.Value) interface{} {
+func (c WebSocket) addEventListener(eventType string, fn func(e js.Value)) func() {
+	f := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
 		fn(args[0])
```
albrow: It's really important to [not block inside of a `js.Func` wrapped function](https://golang.org/pkg/syscall/js/#FuncOf). Doing so will block the JavaScript event loop and prevent any other JavaScript functions from running. You should almost always start a new goroutine inside of `js.FuncOf`.

For example:

```go
	f := js.FuncOf(func(this js.Value, args []js.Value) interface{} {
		go fn(args[0])
		return nil
	})
```

internal/wsjs/wsjs.go
```diff
@@ -119,23 +131,29 @@ func (c *WebSocket) OnMessage(fn func(m MessageEvent)) {
...
 	c.v.Call("close", code, reason)
 	return err
 }
 
-func (c *WebSocket) SendText(v string) (err error) {
+// SendText sends the given string as a text message
+// on the WebSocket.
+func (c WebSocket) SendText(v string) (err error) {
 	defer handleJSError(&err, nil)
 	c.v.Call("send", v)
```
albrow: Do we have good test coverage for the `SendText` method? I've had a lot of trouble with things like this in the past because JavaScript does wonky and unexpected things with text encoding. The string you are trying to send might not end up being the same as what is received on the other end. It might even be better to just remove this method.

websocket_js.go
```diff
@@ -0,0 +1,217 @@
+package websocket // import "nhooyr.io/websocket"
```
albrow: Shouldn't this file have the build tag: `+build js,wasm`?

wsjson/wsjson_js.go
```diff
@@ -0,0 +1,58 @@
+// +build js
```
albrow: Unless you know that a file is compatible with [GopherJS](https://github.com/gopherjs/gopherjs) you usually want to use the build tag `+build js,wasm`. The addition of `wasm` clarifies for the Go compiler that this file is compatible with WebAssembly but tells the GopherJS compiler that the file is not compatible with GopherJS. This [blog post explains the situation in a little more detail](https://medium.com/gopherjs/future-of-gopherjs-and-go-in-the-browser-cc1667468160).

Basically it works like this:

|  Build tag  | Vanilla Go | WebAssembly | GopherJS |
| ---------- | ---------- | -------------- | ---------- |
| (none)       |  :white_check_mark:  | :white_check_mark: | :white_check_mark: |
| `js`            | :x:              |  :white_check_mark: |  :white_check_mark: |
| `!js`           |  :white_check_mark: | :x: | :x: |
| `js,wasm`  | :x: |  :white_check_mark: | :x: |
| `js,!wasm` | :x: |  :x: |  :white_check_mark: |

README.md
```diff
@@ -16,18 +16,18 @@ go get nhooyr.io/websocket
...
 - Minimal and idiomatic API
-- Tiny codebase at 1700 lines
+- Tiny codebase at 2200 lines
 - First class [context.Context](https://blog.golang.org/context) support
 - Thorough tests, fully passes the [autobahn-testsuite](https://github.com/crossbario/autobahn-testsuite)
 - [Zero dependencies](https://godoc.org/nhooyr.io/websocket?imports)
 - JSON and ProtoBuf helpers in the [wsjson](https://godoc.org/nhooyr.io/websocket/wsjson) and [wspb](https://godoc.org/nhooyr.io/websocket/wspb) subpackages
 - Highly optimized by default
 - Concurrent writes out of the box
+- [Complete WASM](https://godoc.org/nhooyr.io/websocket#hdr-WASM) support
```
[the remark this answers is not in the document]
nhooyr: Fixed.

internal/wsjs/wsjs.go
nhooyr: Nice catch, fixed.

internal/wsjs/wsjs.go
nhooyr: We have good test coverage here. See https://github.com/nhooyr/websocket/blob/ef5a88d1828ab3da40b07b9faacf627b0421b329/websocket_js_test.go#L39

websocket_js.go
nhooyr: Files with _js are automatically compiled only for GOOS=js.

wsjson/wsjson_js.go
nhooyr: The library should just work with GopherJS once https://github.com/gopherjs/gopherjs/issues/929 is closed.

nhooyr: It won't even compile on GopherJS versions that don't support Go 1.13 because we use the new wrapping functions in the errors package.

nhooyr: reason being syscall/js is shimmed by GopherJS so that it works the same as Go Wasm.

nhooyr:
> Is there any plan to implement io.ReadWriteCloser on top of this API? that would be really helpful for a variety of use cases.

Already implemented. The websocket.NetConn is compiled into Wasm builds.

nhooyr:
In this PR it wasn't but I made it work in follow up PR #146

internal/wsjs/wsjs.go
albrow: @nhooyr it looks like [`randString` only includes hex characters](https://github.com/nhooyr/websocket/blob/7b4bd30d74307696a93536d863763f627e459c84/assert_test.go#L102)? You should probably test with unicode characters, in particular ones with higher code points. There are some differences in how these are encoded in Go and JavaScript.

nhooyr: That's fair. I'll adjust it.

nhooyr: Fixed in #148

wsjson/wsjson_js.go
albrow: Gotcha. Seems fine if you are not supporting versions older than 1.13.

websocket_js.go
albrow: 👌 Combined with the other thread about not needing `wasm` in the build tags, this should be fine.

PR #150, opened 2019-09-26
Small code changes

cristaloleg:

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -19,6 +19,10 @@ import (
 	"time"
 )
 
+func init() {
+	rand.Seed(time.Now().UnixNano())
+}
+
 // Conn represents a WebSocket connection.
 // All methods may be called concurrently except for Reader, Read
 // and SetReadLimit.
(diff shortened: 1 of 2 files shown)


conn.go
```diff
@@ -838,10 +841,6 @@ func (c *Conn) writeClose(p []byte, cerr error) error {
 	return nil
 }
 
-func init() {
```
nhooyr: I'd prefer this stays here as it's only used with Ping.

conn.go
```diff
@@ -685,9 +689,8 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, opcode opcode, p []byte
 	case <-c.closed:
 		return n, c.closeErr
 	case c.setWriteTimeout <- context.Background():
+		return n, nil
```
nhooyr: Should be reverted, was intentional to match the write timeout code above.

conn.go
```diff
@@ -70,7 +74,7 @@ type Conn struct {
 	activeReader *messageReader
 	// readFrameLock is acquired to read from bw.
 	readFrameLock     chan struct{}
-	readClosed        int64
+	isReadClosed      int64
```
nhooyr: This is 👍

conn.go
cristaloleg: ok

PR #151, opened 2019-09-26
Consolidate status codes

cristaloleg:


frame.go
```diff
@@ -204,31 +204,29 @@ type StatusCode int
 // These codes were retrieved from:
 // https://www.iana.org/assignments/websocket/websocket.xhtml#close-code-number
 const (
-	StatusNormalClosure StatusCode = 1000 + iota
```
nhooyr: This is 👍  since iota does make it harder to tell what number each status code corresponds.

frame.go
```diff
@@ -272,7 +270,7 @@ func parseClosePayload(p []byte) (CloseError, error) {
 // and https://tools.ietf.org/html/rfc6455#section-7.4.1
 func validWireCloseCode(code StatusCode) bool {
 	switch code {
-	case 1004, StatusNoStatusRcvd, StatusAbnormalClosure, StatusTLSHandshake:
+	case StatusReserved, StatusNoStatusRcvd, StatusAbnormalClosure, StatusTLSHandshake:
```
nhooyr: Can you unexport status reserved?

cristaloleg: hm, agree, should be better

nhooyr:
@cristaloleg Could you change the second commit to start with a capital? Then this should be good for merge.

PR #158, opened 2019-10-04
Switch CI to Typescript

nhooyr:
Significantly faster locally since **everything** runs in parallel.

Also much easier to maintain than bash.


nhooyr:
Need to update the contributing docs.

PR #159, opened 2019-10-05
Add websocket.CloseStatus

nhooyr:
Will make it much more convenient and less error prone to access
the StatusCode of a *CloseError.

Related #128 

Precedent for this is in https://godoc.org/gocloud.dev/gcerrors#Code

It's annoying to do the whole xerrors.As dance when you only want to access the status code.

I was only blocked on the name for this but I think `CloseStatus` is nice.

@ammar @coadler

### DIFF
--- a/frame.go
+++ b/frame.go
@@ -253,11 +253,11 @@ func (ce CloseError) Error() string {
 	return fmt.Sprintf("status = %v and reason = %q", ce.Code, ce.Reason)
 }
 
-// CloseStatus is a convenience wrapper around xerrors.As to grab
+// CloseStatus is a convenience wrapper around errors.As to grab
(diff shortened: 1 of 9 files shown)


frame.go
```diff
@@ -252,6 +253,17 @@ func (ce CloseError) Error() string {
...
+// the status code from a *CloseError. If the passed error is nil
+// or not a *CloseError, the returned StatusCode will be -1.
+func CloseStatus(err error) StatusCode {
+	var ce *CloseError
+	if errors.As(err, &ce) {
+		return ce.Code
+	}
+	return -1
+}
+
```
nhooyr: This is the main addition in this PR, the rest is just some cleanup.

nhooyr:
Will mark this experimental for now.

PR #163, opened 2019-10-15
Rewrite with compression support

nhooyr:
https://tools.ietf.org/html/rfc7692


nhooyr:
Memory usage is still a problem but I think this feature can still be useful in some cases and it is standard so might as well implement it given it doesn't add too much to the API.

nhooyr:
Thoughts on the API? @andersfylling @coadler @FZambia

andersfylling:
It's not relevant to my use-cases as I use gzip, I think(?). But looks easy enough to use.

I would add a validation step on Treshold and Level field.

nhooyr:
Also, I noticed Facebook uses per message deflate for their chat, which has very small message sizes. Could either be a oversight or an optimization, not sure. 🤔

nhooyr:
So this now includes a pretty sizeable rewrite of the library.

It was hard to integrate compression due to coupling in the reading/writing implementation and some complexity creep over the last few months so I cleaned up a lot.

nhooyr:
The code in this PR is in a good spot now in case anyone would like to test it out.

nhooyr:
Gonna write a bunch of tests this weekend followed by a serious review and then we finally release :)

PR #170, opened 2019-11-05
Pass on the underlying error for wsjson.Read and wsjson.Write

Jahaja:
This change is to avoid a double error wrap which enables one to easier use the `CloseStatus` helper on error.


nhooyr:
`CloseStatus` should work on the error as is given it uses Go 1.13's errors.As.

See https://blog.golang.org/go1.13-errors

Jahaja:
Whoops. I'm sorry for wasting your time.

PR #177, opened 2019-12-06
Improve handshake to better match RFC

### LINKED ISSUE #173: Return HTTP 426 status code when appropriate
At the moment during a handshake error the server will response with 400 status code for example:

```
HTTP/1.1 400 Bad Request
Content-Length: <content length>
Content-Type: text/plain; charset=utf-8

websocket protocol violation: Connection header "" does not contain Upgrade
```

However, wouldn't it be beneficial to respond with 426 status code when incorrect Connection/Upgrade header is given or the HTTP protocol of the request is bellow 1.1?

```
HTTP/1.1 426 Upgrade Required
Upgrade: WebSocket
Connection: Upgrade
Content-Length: <content length>
Content-Type: text/plain; charset=utf-8

websocket protocol violation: Connection header "" does not contain Upgrade
```

Also I would appreciate if there would be an option to disable the functionality of `Accept` which writes the error to the response and I could handle it "my way".

### LINKED ISSUE #176: Sec-WebSocket-Version missing from response on handshake error
RFC 6455 states the following in [Section 11.3.5. Sec-WebSocket-Version](https://tools.ietf.org/html/rfc6455#section-11.3.5):

>   The |Sec-WebSocket-Version| header field is used in the WebSocket
>   opening handshake.  It is sent from the client to the server to
>   indicate the protocol version of the connection.  This enables
>   servers to correctly interpret the opening handshake and subsequent
>   data being sent from the data, and close the connection if the server
>   cannot interpret that data in a safe manner.  **The |Sec-WebSocket-
>   Version| header field is also sent from the server to the client on
>   WebSocket handshake error, when the version received from the client
>   does not match a version understood by the server.  In such a case,
>   the header field includes the protocol version(s) supported by the
>   server.**

nhooyr:
Closes #176
Closes #173

### DIFF
--- a/handshake.go
+++ b/handshake.go
@@ -55,14 +55,18 @@ func verifyClientRequest(w http.ResponseWriter, r *http.Request) error {
 	}
 
 	if !headerValuesContainsToken(r.Header, "Connection", "Upgrade") {
+		w.Header().Set("Connection", "Upgrade")
+		w.Header().Set("Upgrade", "websocket")
 		err := fmt.Errorf("websocket protocol violation: Connection header %q does not contain Upgrade", r.Header.Get("Connection"))
-		http.Error(w, err.Error(), http.StatusBadRequest)
+		http.Error(w, err.Error(), http.StatusUpgradeRequired)
 		return err
 	}
 
 	if !headerValuesContainsToken(r.Header, "Upgrade", "WebSocket") {
+		w.Header().Set("Connection", "Upgrade")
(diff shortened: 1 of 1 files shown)


nhooyr:
Please review @nerg4l

nhooyr:
Thanks @nerg4l

nerg4l:
Anytime @nhooyr . I will keep looking for improvements.

PR #193, opened 2020-02-20
Ensure connection is closed at all error points

### LINKED ISSUE #191: Seeing failed to get reader and to write msg errors
Hi, I just updated your library to the last version; I tried a basic example with two Websocket connection and get these errors in my Go debug console:

`failed to get reader: received close frame: status = StatusGoingAway and reason = ""`
on page load (when HTML page try to open the Websocket connection).

and:

`write: failed to write msg: failed to close writer: failed to write fin frame: failed to flush: write tcp 127.0.0.1:7070->127.0.0.1:61615: write: broken pipe`
on writing (when my Go code try to send a message to the HTML page).

These errors pop out with Firefox (73) and Chrome (79), but I think they are mostly "warnings" because everything seems to work.
Anyway, I never had these errors in older versions of your library and my frontend code is still the same.
Can you investigate, please?

nhooyr:
Closes #191

### DIFF
--- a/read.go
+++ b/read.go
@@ -304,7 +304,9 @@ func (c *Conn) reader(ctx context.Context) (_ MessageType, _ io.Reader, err erro
 	defer c.readMu.unlock()
 
 	if !c.msgReader.fin {
-		return 0, nil, errors.New("previous message not read to completion")
+		err = errors.New("previous message not read to completion")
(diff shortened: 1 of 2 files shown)


nhooyr:
cc @coadler

coadler:
:+1: lgtm

PR #198, opened 2020-02-26
Add OriginPatterns to AcceptOptions

### LINKED ISSUE #194: How to set Cross domain origin?
in AcceptOptions struct, I found the `InsecureSkipVerify` field and set it equal to true.  but all clients can connect to the WebSocket server, it includes illegal clients.  in [gorilla](https://github.com/gorilla/websocket) have [CheckOrigin](https://github.com/gorilla/websocket/blob/master/server.go#L66). but there is no the same method. (AcceptOrigins and add AcceptInsecureOrigin is removed in the current version).
```

	// InsecureSkipVerify disables Accept's origin verification behaviour. By default,
	// the connection will only be accepted if the request origin is equal to the request
	// host.
	//
	// This is only required if you want javascript served from a different domain
	// to access your WebSocket server.
	//
	// See https://stackoverflow.com/a/37837709/4283659
	//
	// Please ensure you understand the ramifications of enabling this.
	// If used incorrectly your WebSocket server will be open to CSRF attacks.
	InsecureSkipVerify bool

```

@nhooyr  How to set origin like gorilla's` CheckOrigin`?   Looking forward to your reply, appreciate it.  Thanks.

nhooyr:
Closes #194

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -9,10 +9,11 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"net/textproto"
 	"net/url"
-	"strconv"
+	"path/filepath"
 	"strings"
 
 	"nhooyr.io/websocket/internal/errd"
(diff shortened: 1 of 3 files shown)


nhooyr:
Would appreciate your review as well @coadler.

This change allows websocket to match request origins with a list of patterns.

Issue #197, opened 2020-02-26
Export wstest.Pipe for in memory tests


nhooyr:
Would be really cool to have an in memory WebSocket implementation for tests versus starting a net/http test server every test.

nhooyr:
Not sure if this is a good idea anymore with #199 🤔

nhooyr:
Any thoughts @coadler?

nhooyr:
Looks like this: https://github.com/nhooyr/websocket/blob/wstest/internal/test/wstest/pipe.go#L17

nhooyr:
```
// Pipe is used to create an in memory connection
// between two websockets analogous to net.Pipe.
func Pipe(dialOpts *websocket.DialOptions, acceptOpts *websocket.AcceptOptions) (clientConn, serverConn *websocket.Conn) {
```

nhooyr:
If I were to export it, I'd change the signature to:

```
func Pipe() (http.RoundTripper, http.ResponseWriter)
```

Any upgrade requests sent on the round tripper would wait for the given response writer to hijack and accept the WebSocket.

nhooyr:
Not sure if this is a good idea anymore as `httptest` makes it so easy already and I don't think there are enough scenarios where you want in memory web sockets vs abstracting away the transport and passing around messages directly. The only good scenario I can really think of is writing tests for the library which is why it exists.

nhooyr:
Yea, going to close this for now. Only real problem is ensuring graceful shutdown in tests for which #209 is open.

PR #202, opened 2020-02-26
Simplify wstest.Pipe

nhooyr:
Will deal with #197 before merging.


nhooyr:
Will deal with it after.

PR #253, opened 2020-07-04
write.go: Fix deadlock in writeFrame

### LINKED ISSUE #248: Lingering Deadlock when Write and CloseRead called concurrently in 1.8.6
In the [fix for the deadlock introduced in 1.8.5](https://github.com/nhooyr/websocket/commit/03cca81f294e0386e8364ed68c397c3f7163ced4), we have the following:

```
err = c.writeFrameMu.lock(ctx)
if err != nil {
	return 0, err
}
defer c.writeFrameMu.unlock()

// If the state says a close has already been written, we wait until
// the connection is closed and return that error.
//
// However, if the frame being written is a close, that means its the close from
// the state being set so we let it go through.
c.closeMu.Lock()
wroteClose := c.wroteClose
c.closeMu.Unlock()
if wroteClose && opcode != opClose {
	select {
	case <-ctx.Done():
		return 0, ctx.Err()
	case <-c.closed:
		return 0, c.closeErr
	}
}

...
```

If there is a `(*Conn).Write` that passes the `if` condition, then it waits for at least one of two channels to close, while still holding the `writeFrameMu` lock. Since `c.wroteClose` is set before `(*Conn).CloseRead` gets to call `(*Conn).writeFrame` (see `(*Conn).writeClose`), it is possible that `(*Conn).CloseRead` calls `(*Conn).writeFrame` after `(*Conn).Write` gets stuck, so `(*Conn).CloseRead` will get stuck waiting to get `writeFrameMu`, and no progress can be made.

One option would be to unlock `writeFrameMu` immediately after passing the `if` condition. `defer c.writeFrameMu.unlock()` would need to be removed since the mutex implementation doesn't allow for unlocking twice.

nhooyr:
Closes #248

Luckily, due to the 5s timeout on the close handshake, this would have
had very minimal effects on anyone in production.

### DIFF
--- a/write.go
+++ b/write.go
@@ -257,7 +257,6 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, flate bool, opcode opco
 	if err != nil {
 		return 0, err
 	}
-	defer c.writeFrameMu.unlock()
 
 	// If the state says a close has already been written, we wait until
 	// the connection is closed and return that error.
@@ -268,6 +267,7 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, flate bool, opcode opco
 	wroteClose := c.wroteClose
 	c.closeMu.Unlock()
 	if wroteClose && opcode != opClose {
+		c.writeFrameMu.unlock()
 		select {
 		case <-ctx.Done():
 			return 0, ctx.Err()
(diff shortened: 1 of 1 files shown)


nhooyr:
Please review @RTann!

nhooyr:
Thanks for reviewing!

PR #255, opened 2020-07-04
netconn.go: Prevent timer leakage

### LINKED ISSUE #243: NetConn leaks timers
Hey, first of all thanks a lot for `NetConn()`! I have written one myself, but then discovered this library and decided to use it instead.

I have noticed that `netConn.Close()` does not call `Stop()` on `writeTimer` and `readTimer`. I think it should at least if the deadlines have been set. Otherwise these timers are lingering around and that can be an issue if there are a lot of connections.

nhooyr:
Closes #243

### DIFF
--- a/netconn.go
+++ b/netconn.go
@@ -106,6 +106,8 @@ type netConn struct {
 var _ net.Conn = &netConn{}
 
 func (nc *netConn) Close() error {
+	nc.writeTimer.Stop()
+	nc.readTimer.Stop()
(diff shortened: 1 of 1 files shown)


nhooyr:
Please review @ash2k!

ash2k:
Looks good to me, thanks.

nhooyr:
Awesome thanks for reviewing.

PR #258, opened 2020-09-09
Make compression negotiation more lenient

abursavich:
I ran into a server this morning that was being extra-super-duper explicit and specifying `server_max_window_bits=15`, so I started digging into the RFC and the negotiation code... These are the changes I would like to make:

1. Client allows server to specify `server_max_window_bits`: The server may use a smaller sliding window for compression without affecting the client's decompression.
2. Server allows client to specify `server_max_window_bits=15`: Although the server can't support a smaller sliding window at the moment, it doesn't hurt to accept the specification of 15 bits from a client.
3. Server declines unacceptable compression offers without rejecting the request and selects the first acceptable offer: According to the RFC, the client may send multiple offers in its preferential order and the server may accept its preferred offer or decline all of them. Declining a compression offer doesn't mean the connection must be terminated.
4. Removes dead `x-webkit-deflate-frame` code.

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -265,6 +265,8 @@ func acceptDeflate(w http.ResponseWriter, ext websocketExtension, mode Compressi
 		case "server_no_context_takeover":
 			copts.serverNoContextTakeover = true
 			continue
+		case "server_max_window_bits=15":
+			continue
 		}
 
 		if strings.HasPrefix(p, "client_max_window_bits") {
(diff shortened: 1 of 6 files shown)


abursavich:
The RFCs allow for some strange formatting of the extensions. For instance, the parameter values may be quoted (e.g. `server_max_window_bits="15"`). I went ahead and implemented full parsing in a different branch. It's not necessary to fix the specific issue I ran into, but it's here if you're interested: https://github.com/abursavich/websocket/compare/compression...abursavich:extensions

nhooyr:
Thanks for contributing @abursavich! <3

Unfortunately I'm going to be a busy this week until the weekend. I'll try and review these changes then.

Have a great rest of your week!

nhooyr (approved):
Beautiful, sorry for the delay.

nhooyr:
Thanks @abursavich!

PR #259, opened 2020-09-09
Clone options

abursavich:
`opts = &*opts` is a no-op.

See: https://staticcheck.io/docs/checks#SA4001

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -63,6 +63,14 @@ type AcceptOptions struct {
 	CompressionThreshold int
 }
 
+func (opts *AcceptOptions) cloneWithDefaults() *AcceptOptions {
+	var o AcceptOptions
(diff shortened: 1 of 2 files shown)


nhooyr:
Nice catch! Can confirm. I tested it at the time and thought it worked but I guess I got confused.

nhooyr (approved):
Thanks for contributing @abursavich!

icholy:
There are still 2 more instances of this.

```
$ git grep "\&\*
dial.go:        copts = &*copts
internal/test/wstest/pipe.go:   dialOpts = &*dialOpts
```

nhooyr:
Nice catches @icholy, I've made note of those in my TODO and will have a PR up to fix shortly.

PR #261, opened 2020-09-22
Copy HTTP client

### LINKED ISSUE #260: Error should not be returned on non-zero http client timeout
Having a warning about non-zero http client timeout is useful, but it should be a documentation-level warning not an actual error. With the current implementation it's not possible to use the same http client for both standard http requests and websocket events, even when a context with timeout is provided during dialing.

Related to: #67

swithek:
Closes #260

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/base64"
-	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -26,6 +25,7 @@ type DialOptions struct {
 	// HTTPClient is used for the connection.
 	// Its Transport must return writable bodies for WebSocket handshakes.
 	// http.Transport does beginning with Go 1.12.
+	// Non-zero timeout will be ignored, see https://github.com/nhooyr/websocket/issues/67.
 	HTTPClient *http.Client
 
 	// HTTPHeader specifies the HTTP headers included in the handshake request.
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
Thanks for contributing @swithek!

Thoughts on instead of removing the timeout completely, copying the client and removing it in Dial and then using whatever timeout is on the client on the Dial context?

That way things would just work as expected.

swithek:
@nhooyr that's a good point, I will submit an update soon.

dial.go
```diff
@@ -74,7 +73,19 @@ func dial(ctx context.Context, urls string, opts *DialOptions, rand io.Reader) (
 	opts = &*opts
 	if opts.HTTPClient == nil {
 		opts.HTTPClient = http.DefaultClient
+	} else if opts.HTTPClient.Timeout > 0 {
+		var cancel context.CancelFunc
+
+		ctx, cancel = context.WithTimeout(ctx, opts.HTTPClient.Timeout)
+		defer cancel()
+
+		opts.HTTPClient = &http.Client{
```
nhooyr, line 82: Let's do the following in case more fields are added to `*http.Client`:

```
newClient = *opts.HTTPClient
newClient.Timeout = 0
opts.HTTPClient = &newClient
```

nhooyr (changes requested):
Looks great otherwise!

swithek:
@nhooyr is there anything else that needs improving?

nhooyr:
Nope, looks great.

Thanks for contributing @swithek <3

nhooyr (approved):

PR #266, opened 2020-11-25
removing bad newline

Zamiell:
this seems like a pointless newline, correct me if I am wrong.

furthermore, I would recommend using [wsl](https://github.com/bombsimon/wsl) in CI to automatically detect these kinds of errors in the future.
wsl is included in [golangci-lint](https://golangci-lint.run/usage/linters/), which is the defacto golang linter, which it seems like you [don't use currently](https://github.com/nhooyr/websocket/blob/master/ci/lint.sh).

### DIFF
--- a/examples/echo/server.go
+++ b/examples/echo/server.go
@@ -16,7 +16,6 @@ import (
 // It ensures the client speaks the echo subprotocol and
 // only allows one message every 100ms with a 10 message burst.
 type echoServer struct {
-
 	// logf controls where logs are sent.
 	logf func(f string, v ...interface{})
 }


examples/echo/server.go
```diff
@@ -16,7 +16,6 @@ import (
 // It ensures the client speaks the echo subprotocol and
 // only allows one message every 100ms with a 10 message burst.
 type echoServer struct {
-
```
nhooyr, line 19: Great catch!

nhooyr:
> furthermore, I would recommend using wsl in CI to automatically detect these kinds of errors in the future.
wsl is included in golangci-lint, which is the defacto golang linter, which it seems like you don't use currently.

I prefer to avoid small dependencies. idt this is common enough a bug to justify incorporating `wsl`.

I disagree that `golangci-lint` is the defacto go linter.

It's `goimports`, `gofmt`, `golint` and `go vet` all of which we use and are from the Go team.

Zamiell:
1) Well, to be technical, `goimports` and `gofmt` are code formatters, not linters. =p
2) golangci-lint includes `goimports`, `gofmt`, `golint`, `govet`, as well as various other useful linters, like the previously mentioned `wsl`.
3) Note that `golint` is now **officially deprecated** and not very many people use it anymore: https://github.com/golang/go/issues/38968

For my personal Golang projects, I've found a tremendous amount of value in using golanci-lint. If you want to keep it simple and not use it, that's fine - I just thought I would throw it out there for consideration.

golanci-lint has some linters enabled by default and some disabled by default. My typical workflow is that I enable every single linter that it offers. This generates a lot of noise, so then I go through and manually disable the ones that I disagree with, and the ones that are noisy. After that, I am left with a few legitimate issues / bugs in my code, which I clean up, making the whole process worth it. =)

An example yaml config file:
https://github.com/Zamiell/hanabi-live/blob/master/server/.golangci.yml

nhooyr:
> Well, to be technical, goimports and gofmt are code formatters, not linters. =p

Fair but the line has grown ever thin. code formatting isn't far from automated linting imo.

> Note that golint is now officially deprecated and not very many people use it anymore: golang/go#38968

That's wild and definitely puts a dent in `golint` for me. I've found it extremely useful, I don't really understand the motivation for removing it in that thread and fully agree with https://github.com/golang/go/issues/38968#issuecomment-634970874

> For my personal Golang projects, I've found a tremendous amount of value in using golanci-lint. If you want to keep it simple and not use it, that's fine - I just thought I would throw it out there for consideration.

Yea I appreciate the thoughts and discussion but for now I'm definitely in favour of keeping things as minimal as possible.

Perhaps I'll switch the library to [Staticcheck](https://github.com/dominikh/go-tools) which seems like it's highly recommended from that thread and a sort of successor to golint.

PR #273, opened 2020-12-21
Preserve client case for selected subprotocol

egorbunov:
WebSocket implementation in chromium rejects
ws connections in case server side responses with
`Sec-WebSocket-Protocol` header which value does
not match one of the values sent in the same header by client

In case you use subprotocols to send case-sensitive text
(for example, auth tokens) you end up with inability
to establish websocket connection from chromium.

#### How to reproduce:

Initialize websocket like this:

```go
websocketOptions := &websocket.AcceptOptions{}
if subProtocol := r.Header.Get("Sec-Websocket-Protocol"); subProtocol != "" {
	websocketOptions.Subprotocols = []string{subProtocol}
}
...
ws, err := websocket.Accept(w, r, websocketOptions)
...
```

After that try to establish connection from javascript in chromium (my current version is `87.0.4280.88`):

```js
var ws = new WebSocket(url, "ToKeN")
```

You will get next error:

```
WebSocket connection to '...' failed: Error during WebSocket handshake: 'Sec-WebSocket-Protocol' header value 'token' in response does not match any of sent values
```

So this PR fixes that behavior.

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -159,13 +159,13 @@ func verifyClientRequest(w http.ResponseWriter, r *http.Request) (errCode int, _
 		return http.StatusUpgradeRequired, fmt.Errorf("WebSocket protocol violation: handshake request must be at least HTTP/1.1: %q", r.Proto)
 	}
 
-	if !headerContainsToken(r.Header, "Connection", "Upgrade") {
+	if !headerContainsTokenIgnoreCase(r.Header, "Connection", "Upgrade") {
 		w.Header().Set("Connection", "Upgrade")
 		w.Header().Set("Upgrade", "websocket")
 		return http.StatusUpgradeRequired, fmt.Errorf("WebSocket protocol violation: Connection header %q does not contain Upgrade", r.Header.Get("Connection"))
 	}
 
-	if !headerContainsToken(r.Header, "Upgrade", "websocket") {
+	if !headerContainsTokenIgnoreCase(r.Header, "Upgrade", "websocket") {
 		w.Header().Set("Connection", "Upgrade")
 		w.Header().Set("Upgrade", "websocket")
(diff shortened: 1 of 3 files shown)


accept.go
```diff
@@ -349,12 +349,18 @@ func websocketExtensions(h http.Header) []websocketExtension {
...
+}
+
+func headerTokensExt(h http.Header, key string, lower bool) []string {
 	key = textproto.CanonicalMIMEHeaderKey(key)
 	var tokens []string
 	for _, v := range h[key] {
 		v = strings.TrimSpace(v)
 		for _, t := range strings.Split(v, ",") {
-			t = strings.ToLower(t)
+			if lower {
```
nhooyr, line 361: Perhaps we shouldn't do a `ToLower` here at all and instead rely on the caller handling case correctly?

nhooyr: I see this code is also going to cause trouble for `Sec-WebSocket-Extensions`. In general, I don't think HTTP header values are case insensitive so I think I made a mistake here.

nhooyr: I think `headerContainsToken` should be the one modified to allow optional case insensitivity as that's the function used to check `Connect` and `Upgrade` whose values are case insensitive. And then this function should be changed to not modify case at all.

nhooyr (commented):
Wonderful catch, thank you @egorbunov <3

accept.go
egorbunov: Rewrote the fix. Decided to rename `headerContainsToken` to `headerContainsTokenIgnoreCase` because this function is only used to deal with `Connect` and `Upgrade` headers.

nhooyr (approved):
Lovely!! Thanks again @egorbunov 🚀

PR #278, opened 2021-01-12
Addition of options to allow loading of custom certificates

dcam0050:
Addition of an Options field of type map[string]interface{} to DialOptions which is used in the constructor for the js websocket instance and enables passing custom options like custom certificates.

### DIFF
--- a/internal/wsjs/wsjs_js.go
+++ b/internal/wsjs/wsjs_js.go
@@ -27,7 +27,7 @@ func handleJSError(err *error, onErr func()) {
 }
 
 // New is a wrapper around the javascript WebSocket constructor.
-func New(url string, protocols []string) (c WebSocket, err error) {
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
Thanks for contributing @dcam0050 <3

This is interesting, I didn't know you could do this. Can you link some docs and document this on the `Options` you added?

dcam0050:
After further investigation it seems like this is an option that is only available when using NodeJS as the javascript interpreter. This is the npm package in was using: https://www.npmjs.com/package/ws

The browser api for Websocket on the other hand does not allow for the use of custom certificates at all. That said though, when running this change with Vanillajs, there was no side effect except for lack of feedback saying that Options were ignored.

nhooyr:
I think this is better solved by #373 then as you could override the transport and not use the browser API as the library assumes by default to pass whatever options you want.

PR #299, opened 2021-04-24
Returns real remote and local address instead mocked

photostorm:
Currently, RemoteAddr() and LocalAddr() are mocked. This change allow the real RemoteAddr() and LocalAddr() to be used.

### DIFF
--- a/netconn_notjs.go
+++ b/netconn_notjs.go
@@ -1,0 +1,20 @@
+//go:build !js
+// +build !js
+
+package websocket
+
+import "net"
+
+func (nc *netConn) RemoteAddr() net.Addr {
+	if unc, ok := nc.c.rwc.(net.Conn); ok {
+		return unc.RemoteAddr()
+	}
+	return websocketAddr{}
+}
+
+func (nc *netConn) LocalAddr() net.Addr {
+	if unc, ok := nc.c.rwc.(net.Conn); ok {
+		return unc.LocalAddr()
+	}
+	return websocketAddr{}
+}
--- a/conn_test.go
+++ b/conn_test.go
@@ -155,8 +155,8 @@ func TestConn(t *testing.T) {
 		n1.SetDeadline(time.Time{})
 
 		assert.Equal(t, "remote addr", n1.RemoteAddr(), n1.LocalAddr())
-		assert.Equal(t, "remote addr string", "websocket/unknown-addr", n1.RemoteAddr().String())
-		assert.Equal(t, "remote addr network", "websocket", n1.RemoteAddr().Network())
+		assert.Equal(t, "remote addr string", "pipe", n1.RemoteAddr().String())
+		assert.Equal(t, "remote addr network", "pipe", n1.RemoteAddr().Network())
 
 		errs := xsync.Go(func() error {
 			_, err := n2.Write([]byte("hello"))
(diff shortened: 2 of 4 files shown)


ZackaryWelch:
I would like to get this in if possible. It's the only thing this library lacks from gorilla/websocket.

netconn_notjs.go
```diff
@@ -0,0 +1,26 @@
...
+
+import "net"
+
+type LocalRemoteAddr interface {
+	RemoteAddr() net.Addr
+	LocalAddr() net.Addr
+}
+
+func (c *netConn) RemoteAddr() net.Addr {
+	if ra, ok := c.c.rwc.(LocalRemoteAddr); ok {
```
nhooyr, line 13: Not sure how this could work? Have you tested this? `c.c.rwc` is not a `net.Conn` and so doesn't have these methods you're checking for.

nhooyr: It only works in tests because the tests don't use a real http transport/server, they're using my pipe abstraction. See https://github.com/nhooyr/websocket/blob/master/internal/test/wstest/pipe.go

nhooyr:
Going to close as I'm assuming this approach won't work. Leaving the issue open at #327 though so I can figure something out here.

netconn_notjs.go
photostorm: It works fine. Its check the read/writer also RemoteAddr() and LocalAddr() defined.

photostorm: Here is example of something similar to what I did:
https://github.com/golang/go/blob/master/src/io/io.go#L408

nhooyr: Have you tested this in production? There's no RemoteAddr or LocalAddr on the bufio.ReadWriter returned to us from the transport.

photostorm: This the echo exmaple.
Screen Shot 2023-03-08 at 12 10 06 PM

nhooyr: Ok now I'm super confused I'll have to look into how that's possible lol

nhooyr:
Ah I understand, it works server side but would not work with connections from `websocket.Dial` because there we only get a `io.ReadWriteCloser` from `net/http`.

photostorm:
What is being store with io.ReadWriterClose when using the websocket.Dial? io.ReadWriterClose is just common abstraction. If the io.ReadWriterClose is being used with net.Conn then we know the local address and remote address methods exists. Just because we see io.ReadWriterClose when looking at the code does not hide the ability to access other methods if what being stored there during runtime supports these methods. Can you provide any screenshots of what is being stored there?

nhooyr:
I don't fully remember the type but it's not a net.Conn. Grep the net/http.Transport code for protocol upgrade and you'll find it. I'm heading out right now otherwise I'd link you.

photostorm:
I will do exploring with that when I get some time. 

Thanks for your responses.

nhooyr:
See https://github.com/nhooyr/websocket/blob/14fb98eba64eeb5e9d06a88b98c47ae924ac82b4/dial.go#L128

We don't get a `net.Conn` from `net/http` when dialing, we only get a `io.ReadWriteCloser`.

nhooyr:
Thanks @photostorm

PR #306, opened 2021-06-11
close pong channel after ping

qcha0:
should close pong channel after ping


nhooyr (commented):
closing channels in Go is not done for resource cleanup, it's to signal no more messages to the peer.

nhooyr:
Here such signalling isn't necessary.

Issue #246, opened 2020-05-28
Add callbacks to log when Ping/Pong is received


adrianlungu:
Hello,

From what I've seen, the `pong` is automatically handled internally in the library, however, in some specific cases it could be useful to run a function when the `pong` comes back (i.e. measuring latency, logging, etc.) 

Think it would be ok to add a handler function that is called when the `pong` is handled, which can be set through an exposed function ?

nhooyr:
Hi!

So this was intentional to avoid a callback based API. In order to measure latency, you just record the time when you call Ping and then when it returns and the difference is the ping pong. For logging you just log after it returns as well.

I’ll be adding an example soon to make using Ping less confusing. See #227

adrianlungu:
Hello! 

Thanks for the quick answer!

So basically the `ping-pong` is done synchronously ?

But then, how can I log the `pong` that is done towards the server when the library receives a `ping` and handles it through `handleControl` internally ?

nhooyr:
> So basically the ping-pong is done synchronously ?

Yes, see https://github.com/nhooyr/websocket/commit/169521697c04f5b5a06b3da51bf4cad56884d2b6#diff-7101a3ea121adf8d5b0acb5a2c520ca3R138-R162

> But then, how can I log the pong that is done towards the server when the library receives a ping and handles it through handleControl internally ?

You can't. Is there a reason you need to do that?

adrianlungu:
> Yes, see 1695216#diff-7101a3ea121adf8d5b0acb5a2c520ca3R138-R162

Thanks! 

Apart from logging, 

> You can't. Is there a reason you need to do that?

I've been trying to debug various scenarios in which I'm using web sockets, as well as stress testing, and it would be helpful to catch the information.

I also have a scenario in which there's more information present in the `ping-pong` process and, while I know it's not idiomatic, being able to log that could be useful.

nhooyr:
> I also have a scenario in which there's more information present in the ping-pong process and, while I know it's not idiomatic, being able to log that could be useful.

That's a fair use case I didn't consider. Unfortunately I can't think of any good API to expose it. Using callbacks would be unfortunate and the stdlib http2 library also doesn't expose this sort of API either, instead favouring for a blocking Ping like us.

For now, you can use [gohack](https://github.com/rogpeppe/gohack) and add a log into the library itself in handleControl.

adrianlungu:
Gohack sounds interesting, I'll look into that, thanks!

[...]

PR #307, opened 2021-06-11
support set ping handler and pong handler

qcha0:
first of all, thanks this good project, we are preparing a tool based on this project.

Sometimes i need do something while receive ping or pong message.
So I added this feature, is it alright

### DIFF
--- a/conn_notjs.go
+++ b/conn_notjs.go
@@ -59,6 +59,8 @@ type Conn struct {
 	pingCounter   int32
 	activePingsMu sync.Mutex
 	activePings   map[string]chan<- struct{}
+	pingHandler   func(ctx context.Context, p []byte) error
+	pongHandler   func(ctx context.Context, p []byte) error
 }
 
 type connConfig struct {
@@ -89,6 +91,9 @@ func newConn(cfg connConfig) *Conn {
 		closed:      make(chan struct{}),
 		activePings: make(map[string]chan<- struct{}),
 	}
+	// set default ping, pong handler
+	c.SetPingHandler(nil)
+	c.SetPongHandler(nil)
 
 	c.readMu = newMu(c)
 	c.writeFrameMu = newMu(c)
--- a/conn_test.go
+++ b/conn_test.go
@@ -103,6 +103,51 @@ func TestConn(t *testing.T) {
 		assert.Contains(t, err, "failed to wait for pong")
 	})
 
+	t.Run("pingHandler", func(t *testing.T) {
+		tt, c1, c2 := newConnTest(t, nil, nil)
+		defer tt.cleanup()
+
+		var count int
+		c2.SetPingHandler(func(context.Context, []byte) error {
+			count++
+			return nil
+		})
+
+		c1.CloseRead(tt.ctx)
+		c2.CloseRead(tt.ctx)
+
+		for i := 0; i < 10; i++ {
+			err := c1.Ping(tt.ctx)
+			assert.Success(t, err)
+		}
+
+		err := c1.Close(websocket.StatusNormalClosure, "")
+		assert.Success(t, err)
+		assert.Equal(t, "count", 10, count)
+	})
+
+	t.Run("pongHandler", func(t *testing.T) {
+		tt, c1, c2 := newConnTest(t, nil, nil)
+		defer tt.cleanup()
+
+		var count int
+		c1.SetPongHandler(func(context.Context, []byte) error {
+			count++
+			return nil
+		})
+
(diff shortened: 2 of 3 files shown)


nhooyr (commented):
Thanks for implementing this @qcha0 

Related issue is #246

What's your usecase out of curiosity for needing to instrument ping/pongs?

qcha0:
just like i want to count the number of ping，or set close the connection if there is no ping message whithin two cycle.

schmidtw:
@nhooyr I need pings/pongs at the application layer for an app we're building a replacement for.  I know work is being done on the `dev` branch, but adding this one feature to a trusted, working code base prevents me from needing to fork the repo for just this one feature we're trying to ship in a week.

nhooyr:
@schmidtw Sorry I won't get to this for a few weeks still. There's another change I wanted to make to let people handle close frames too. What feature are you trying to ship that relies on this?

schmidtw:
I have a protocol that depends on sending & receiving websocket pings and pongs.  Each side of the connection expect a ping/pong within an interval & if either doesn't get the ping or pong from the other side the connection is determined to be broken & application logic notes the reason, closes down immediately & the client side will reconnect.  The pings and pongs are reported as metrics as well.  This is the project: https://xmidt.io/

Since this is weeks out & behind other PRs and the `dev` branch, it makes sense for me to copy the code, apply this patch & use it locally in my project.  When these features are available upstream I can delete the code from the project & move back to this code.  I'm not interested in staying on a forked copy for any longer than is needed.

nhooyr:
Makes sense. Yes, please fork the code for now.

igolaizola:
We have a similar use case where we need to send a Ping if the device hasn’t sent one within a specified time frame before closing the connection for inactivity, and this PR looks like it could address our requirements perfectly. Could you please share the current status?

We’re planning to migrate from gorilla/websocket to this library, but this feature is the missing piece that’s currently blocking us from completing the migration. We truly appreciate all the effort that’s gone into this project and would prefer to avoid maintaining a fork, as it can be challenging and time-consuming to manage.

Thanks again for all the great work on this library! :raised_hands:

igolaizola:
Since @coder now maintains this project, I'm pinging @mafredri to see if they can take a look. Thanks!

mafredri:
@igolaizola for status, I can share that nobody is currently working on this and this PR can't be merged as-is. We can look into getting this functionality merged, however, I see at least a few issues we'd want to address:

1. Assigning the handler is racy, ideally this would be done as an option when creating the websocket, or we need to protect the handlers via mutex.
2. It's not possible to control pongs, they are sent unconditionally and without delay
3. The custom ping handler can return an error, but it's unclear why or what should happen if it does

Questions that spring to mind are:

a. Do users need to modify handlers at runtime?
b. Do users need more control or are they satisfied with observability?

From discussion here and in #246, it seems like observability is the main key which answers 2 and b.

I don't see a use-case for 1 and a, though, and if it indeed is a use-case, perhaps they'd need the ability to assign multiple handlers (register/deregister style).

I'd be happy if people chimed in here as it'll help lock-down the feature set, cheers.

igolaizola:
Thank you for your quick reply!
I agree that this PR has some issues in its current state.

> a. Do users need to modify handlers at runtime?

I believe that setting handlers during creation is sufficiently flexible. Complex logic can always be encapsulated within a custom handler, and I can't think of a use case where this wouldn’t be adequate.

> b. Do users need more control or are they satisfied with observability?

In our scenario, battery-powered devices act as clients, and our cloud serves as the server. This setup requires that pings are not sent automatically because the devices are responsible for initiating them. Since we’re dealing with battery-constrained devices, optimizing resource usage is critical.
Additionally, we need the ability to trigger a ping to the device at specific moments. If the server doesn’t receive a ping from the device within X seconds, it sends a ping to determine whether the connection is still active. This is necessary because packets can sometimes be lost, meaning a ping from the device might not reach the server.

Would you be open to reviewing a new PR that offers more flexible handling of Ping/Pong messages? The premise would be that these handlers are configured only during startup, and ideally the default functionality remains unchanged to avoid breaking existing implementations.

mafredri:
That's an interesting use-case @igolaizola, thanks for sharing it.

> Would you be open to reviewing a new PR that offers more flexible handling of Ping/Pong messages? The premise would be that these handlers are configured only during startup, and ideally the default functionality remains unchanged to avoid breaking existing implementations.

Sounds great, I'm happy to review!

igolaizola:
@mafredri I have created a new PR here: https://github.com/coder/websocket/pull/509

PR #324, opened 2021-12-10
Full-duplex chat example (with /publish POST endpoint)

SlIdE42:


examples/chat/chat.go
```diff
@@ -76,6 +76,7 @@ func (cs *chatServer) subscribeHandler(w http.ResponseWriter, r *http.Request) {
 	}
 	defer c.Close(websocket.StatusInternalError, "")
 
+	c.SetReadLimit(8192)
```
nhooyr, line 79: there's already a sufficient default read limit.

examples/chat/chat.go
```diff
@@ -129,6 +128,22 @@ func (cs *chatServer) subscribe(ctx context.Context, c *websocket.Conn) error {
 	cs.addSubscriber(s)
 	defer cs.deleteSubscriber(s)
 
+	go func() {
```
nhooyr, line 131: without the `ctx` from `closeread`, you'll have to make your own or `defer close(s.msgs)` here.

nhooyr (changes requested):
thanks for your contribution @SlIdE42 but I have quite a few changes planned for this example as part of #324 and it will be easier for me to do them myself.

PR #329, opened 2022-02-20
Fix close writing even if context is cancelled

kylecarbs:
In an HTTP handler where the context is used for cleanup, a goroutine would leak temporarily. By using the write context, it should block the WebSocket from sending the close frame at all if cancelled.


kylecarbs:
I needed to defer `Close()` in my HTTP handler, and this issue no longer exists!

PR #334, opened 2022-03-21
refactor: run gofmt -w on all files

paralin:
Run gofmt to bring the files up to the latest go 1.17 format.


nhooyr:
I appreciate it @paralin but I'll do this as part of #359 soon.

PR #256, opened 2020-07-04
v1.8.8

nhooyr:
Will document release notes in this PR soon.

PR #336, opened 2022-04-28
Extend DialOptions to allow Host header override

bendiscz:
The standard `http.Request.Host` field allows to override the `Host` header that is sent with the HTTP request (`URL.Host` is used by default). Such option can be useful in various scenarios, for example when running behind some proxy server.

This patch extends the `DialOptions` struct with `Host` field that basically does the same.

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -30,6 +30,10 @@ type DialOptions struct {
 	// HTTPHeader specifies the HTTP headers included in the handshake request.
 	HTTPHeader http.Header
 
+	// Host optionally overrides the Host HTTP header to send. If empty, the value
+	// of URL.Host will be used.
+	Host string
+
 	// Subprotocols lists the WebSocket subprotocols to negotiate with the server.
 	Subprotocols []string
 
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
Couldn't you use `Host` in `HTTPHeader`?

bendiscz:
> Couldn't you use `Host` in `HTTPHeader`?

That does not work unfortunately. 'http.Client.Do` ignores `Host` in `HTTPHeader`, you need to set the `http.Request.Host` field if you want to override host from `http.Request.URL`.

nhooyr:
Sounds good will get this in then after #256

nhooyr:
Or you know what we could actually check for `Host` in `HTTPHeader` and special case it to apply to `http.Request.Host`. The API in `net/http` isn't something we should follow to the tee anyway as it's hodgepodge.

bendiscz:
Checking for `Host` in `HTTPHeader` works too, I just wanted to keep the API similar to net/http... But that is certainly not required :-)

Should I change the patch?

nhooyr:
Go for it and thank you for your contribution. It's still going to take me a while to merge as I need to take care of #256

Issue #341, opened 2022-06-02
document http client must use context timeout and not timeout on struct


yihexi:
when I set  http.DefaultClient.Timeout = time.Second * 10, websocket.Dial will get an error "failed to WebSocket dial: response body is not a io.ReadWriteCloser: *http.cancelTimerBody"

verson: nhooyr.io/websocket v1.8.7

this is because in the dial.go line 128:	rwc, ok := respBody.(io.ReadWriteCloser), it convert respBody to an io.ReadWriteCloser, I think it should be rwc, ok := respBody.(io.ReadCloser). 

in general, respBody is a *http.readWriteCloserBody, but if http.DefaultClient.Timeout is set, it will be a *http.cancelTimerBody, which is just an io.ReadCloser.

nhooyr:
Ah right, I'll document this.

mateusfmello:
@nhooyr 

Was this implemented in the project?

I'm having the same problem here.

nhooyr:
Nothing to implement, just an unfortunate quirk of `*http.Client`. Please use the context to pass in a timeout.

mateusfmello:
Looking at the readme example, I switched to `context.WithTimeout`, but the error continues

nhooyr:
Ah right yea so I'll change the code here to remove the Timeout from `*http.Client` but right now you have to make sure Timeout is `0` on whatever `*http.Client` you use with `websocket.Dial`.

mateusfmello:
I believe that nothing needs to be done, as `context.WithCancel(context.Background())` works perfectly too, I believe the problem in my code is something else, I'm rewriting it.

The bug occurred after closing the connection and reopening

nhooyr:
Oh interesting, I'll definitely test to make sure.

nhooyr:
Fixed in dev.

PR #342, opened 2022-06-02
fix bug: issue #341

yihexi:
fix bug :
Dial got error "failed to WebSocket dial: response body is not a io.ReadWriteCloser: *http.cancelTimerBody" when set http.DefaultClient.Timeout #341

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -1,3 +1,4 @@
+//go:build !js
 // +build !js
 
 package websocket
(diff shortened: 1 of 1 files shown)


MajorLettuce:
Just got the same error when I was using http.DefaultClient prior to dialing a websocket connection.
I solved it by creating my own *http.Client instance and setting Timeout there:

```go
httpClient := &http.Client{
  Timeout: 10 * time.Second,
}
```

nhooyr:
@MajorLettuce That shouldn't have worked either though. DefaultClient is a `*http.Client`.

@yihexi This code won't compile, we need to be able to write to the connection.

PR #344, opened 2022-06-23
Update README.md

Critters:
The lack of panics and error catching on ListenAndServe caught me out. This fixes it for future noobs.


nhooyr:
I understand it could have been confusing if you're new to Go but the omissions were intentional for brevity. `panic(err)` is not an appropriate error handling method here. You'd want to log and return but that makes it significantly more verbose.

PR #360, opened 2022-12-20
Reject invalid "Sec-WebSocket-Key" headers from clients

Emyrk:
Client "Sec-WebSocket-Key" should be a valid 16 byte base64 encoded nonce. If the header is not valid, the server should reject the client.

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -1,4 +1,3 @@
-//go:build !js
 // +build !js
 
 package websocket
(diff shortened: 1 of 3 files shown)


Emyrk:
@nhooyr this was brought up by a security consultant. Would be great to get this in so we do not need to handle this at a layer above the websocket library.

nhooyr (commented):
Will get this in for v1.8.8. But I need to mention that it was entirely intentional on my end as it's a useless check mandated by the RFC. This isn't a security issue whatsoever and so I decided to ignore the RFC here and not waste CPU cycles on checking. But I can understand that as this library gets more usage it's important to stick to the letter of the RFC to satisfy consultants and process.

Also for anyone testing their websocket implementation against my own, it's good to error here as other libraries may not be flexible.

nhooyr (approved):

nhooyr:
Will fix the merge conflicts myself.

nhooyr (approved):

Issue #297, opened 2021-04-15
Remove third party test dependencies


yinzara:
The github.com/gin-gonic/gin dependency has a http response splitting vulnerability in versions less than 1.7.0:
https://github.com/gin-gonic/gin/pull/2632

Upgrading the dependency to 1.7.0 or higher will fix this vulnerability (current version is 1.7.1)

nhooyr:
Hi @yinzara we don't actually use gin anywhere other than in tests but I'll be sure to do this in the next release.

nhooyr:
And apologies for the delayed response.

jmichalak-fluxninja:
@nhooyr Any update on this? Github alerts me with https://github.com/advisories/GHSA-h395-qcrw-5vmq

nhooyr:
> https://gophers.slack.com/archives/C9BMAAFFB/p1691163449401899

@prochac can you copy and paste the contents of the link? I'm not a member of the Gophers Slack.

nhooyr:
Done in dev.

PR #361, opened 2023-01-13
Versions

nagachetan:


nhooyr:
I cannot merge this in due to other pending changes. See #297 and #256

PR #365, opened 2023-01-30
Fix dial panic when ctx is nil

guseggert:
When the ctx is nil, http.NewRequestWithContext returns a "net/http: nil Context" error and a nil request. In this case, the dial function panics because it assumes the req is never nil. This checks the returning error and returns it, so that callers get an error instead of a panic in that scenario.

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -159,7 +159,7 @@ func handshakeRequest(ctx context.Context, urls string, opts *DialOptions, copts
 
 	req, err := http.NewRequestWithContext(ctx, "GET", u.String(), nil)
 	if err != nil {
-		return nil, fmt.Errorf("failed to build HTTP request: %w", err)
+		return nil, fmt.Errorf("http.NewRequestWithContext failed: %w", err)
 	}
 	req.Header = opts.HTTPHeader.Clone()
 	req.Header.Set("Connection", "Upgrade")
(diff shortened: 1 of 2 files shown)


dial.go
```diff
@@ -157,7 +157,10 @@ func handshakeRequest(ctx context.Context, urls string, opts *DialOptions, copts
 		return nil, fmt.Errorf("unexpected url scheme: %q", u.Scheme)
 	}
 
-	req, _ := http.NewRequestWithContext(ctx, "GET", u.String(), nil)
+	req, err := http.NewRequestWithContext(ctx, "GET", u.String(), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build HTTP request: %w", err)
```
nhooyr, line 162:
```suggestion
		return nil, fmt.Errorf("http.NewRequestWithContext failed: %w", err)
```

nhooyr (changes requested):
Thanks @guseggert

BigLep:
@nhooyr : is this going to get merged?

elimisteve:
LGTM :+1:

nhooyr:
I'm not sure about this anymore. ctx should never be nil. You should just use `context.Background` instead.

@BigLep @elimisteve why are you guys running into this?

nhooyr:
Yea there doesn't seem to be a consistent convention on this even in the go stdlib. net.DialContext for example panics on a nil context. https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/net/dial.go;l=456

Seems right to me to panic. I'm happy to incorporate the missing err check on `http.NewRequestWithContext` though.

nhooyr (approved):

nhooyr:
Thanks @guseggert

PR #366, opened 2023-01-31
go.mod: bump gin to v1.7.0 to fix CVE-2020-28483

Xe:
Signed-off-by: Xe Iaso <xe@tailscale.com>


marcus-vw:
Can we merge the PR. The changes are reasonable.

komuw:
If we run https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck on this repository, it does not list it as been affected by `CVE-2020-28483`. And yes, that CVE does exist in its database; https://pkg.go.dev/vuln/GO-2021-0052 .
There are a lot of vuln. scanners out there that raise an alert just because a dependency exists(and at a certain version) without analysing whether the actual defect is exercised in the codebase.

nhooyr:
See #297

PR #345, opened 2022-07-05
github.com/gin-gonic/gin version bump

sdrobov:
Dependabot alerts, that github.com/gin-gonic/gin < 1.7.0 has critical vulnerability, so I bumped version to latest atm.

PR #319, opened 2021-10-02
Bump github.com/gin-gonic/gin from 1.6.3 to 1.7.4

kirill-scherba:
issue #318

PR #369, opened 2023-02-14
remove test gin dependency

tychoish:


tychoish:
sup nhooyr! :)

tychoish:
I think this is probably a better fix than:
- #345 
- #366 
- #319

nhooyr:
@tychoish Hey Sam!

This test intentionally imports and tests gin. There's other end to end tests (or they might be only in the dev branch).

Gin is a terrible piece of software but widely used and it changes the behaviour of `http.ResponseWriter` which is why I felt the need to test against it directly.

But given the stupidity of go modules including test modules in downstream dependencies, I'll move the gin test into a different repository entirely eventually.

For now, master can't have anything merged into until I deal with #256

nhooyr:
> That makes sense. I've had some luck with having x/<foo> packages with their own sub-modules to isolate dependencies for cases kinda like this. It works... fine (with the annoying side effect of making go test ./... skip things you wish it didn't, but you can get around it with a little find/xargs magic). Anyway, seems reasonable.

Good to know it works.

> Is there something technical blocking https://github.com/nhooyr/websocket/pull/256, or just time? Is there some work I (or someone) could do to help out?

Failing tests. I appreciate your initative but it's something I need to review and fix myself.

PR #393, opened 2023-05-18
Add info about Gorilla WebSocket being Archived

KunalSin9h:


nhooyr:
Hi @KunalSin9h I appreciate your contribution but it looks like gorilla is now unarchived. Sorry for my delay.

Issue #298, opened 2021-04-16
Ping sometimes causes Close to time out


univerio:
Minimal reproduction:

```go
package websocket_test

import (
	"context"
	"testing"
	"time"

	"github.com/stretchr/testify/assert"
	"nhooyr.io/websocket"
	"nhooyr.io/websocket/internal/test/wstest"
)

func TestWeirdTimeout(t *testing.T) {
	sConn, cConn := wstest.Pipe(nil, nil)
	sConn.CloseRead(context.Background())
	cConn.CloseRead(context.Background())
	go func() {
		for range time.Tick(10 * time.Millisecond) {
			if err := sConn.Ping(context.Background()); err != nil {
				return
			}
		}
	}()

	time.Sleep(30 * time.Millisecond)
	assert.NoError(t, sConn.Close(websocket.StatusNormalClosure, ""))
}
```

This seems to non-deterministically fail with the documented 5s timeout:

```
$ go test -run TestWeirdTimeout -count 100 -failfast -v
=== RUN   TestWeirdTimeout
--- PASS: TestWeirdTimeout (0.03s)
=== RUN   TestWeirdTimeout
--- PASS: TestWeirdTimeout (0.03s)
=== RUN   TestWeirdTimeout
--- PASS: TestWeirdTimeout (0.03s)
=== RUN   TestWeirdTimeout
    foo_test.go:26:
        	Error Trace:	foo_test.go:26
        	Error:      	Received unexpected error:
        	            	failed to close WebSocket: failed to write control frame opClose: failed to write frame: context deadline exceeded
        	Test:       	TestWeirdTimeout
--- FAIL: TestWeirdTimeout (5.03s)
FAIL
exit status 1
FAIL	nhooyr.io/websocket	5.309s
```

Is it incorrect to call `Close()` concurrently with `Ping()`?

nhooyr:
Hi @univerio 

So I'm not 100% sure but I believe this is expected behaviour as the Ping being every 10 milliseconds is effectively hogging the write lock.

Instead of a ticker, try using a timer that is reset after every ping/pong. I'd bet it happens far less often if never as there would be a guaranteed 10ms interval between pings/pongs then.

I'll be sure to test prior to the next release myself.

And apologies for the delayed response!

univerio:
I made it 10ms to make it fail faster, but it happens with 1s tick and 3s closure delay as well.

[...]

PR #394, opened 2023-05-25
Avoid handling ping after close frame has been sent

bhallionOhbibi:
Fixes https://github.com/nhooyr/websocket/issues/298

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -553,3 +553,26 @@ func assertClose(tb testing.TB, c *websocket.Conn) {
 	err := c.Close(websocket.StatusNormalClosure, "")
 	assert.Success(tb, err)
 }
+
+func TestConcurrentClosePing(t *testing.T) {
+	t.Parallel()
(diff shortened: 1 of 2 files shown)


bhallionOhbibi:
Any chance you have time to take a quick look at this ?
This fix is quite simple :D

nhooyr:
Unfortunately this doesn't work. writeFrame already has a check like this built in so it makes sense it doesn't work. There is definitely a bug in #298, but need to dig deeper to figure out what.

See the test I added to your commit, it still fails.

nhooyr:
See https://github.com/nhooyr/websocket/commit/e4879ab74e5dd045a4ef5707f2c542b9cf4a4321 for the test

nhooyr:
Fixed in dev, see https://github.com/nhooyr/websocket/commit/28c670953e8a6c6ecf147f89ac0085ac6510999e

PR #412, opened 2023-10-25
Change protocol in docs and logs from http to ws

KunalSin9h:
the websocket protocol require the protocol on the url to be `ws` not `http`, then we must log the url which is for websocket.


nhooyr (approved):
Good catch.

PR #428, opened 2023-12-19
Provide a simple callback function for when you need to know a ping came in.

schmidtw:
This is a simple approach to providing users with a relatively easy and safe way to be notified of when a `ping` comes from the server to the client.  It doesn't allow for any action to be take aside from letting 3rd party code know something happened.

This should address #246.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -79,6 +79,8 @@ type Conn struct {
 	pingCounter   int32
 	activePingsMu sync.Mutex
 	activePings   map[string]chan<- struct{}
+
+	pingCallback func()
 }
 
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
Thanks for contributing @schmidtw but there's already #307 and I believe it has a better API with the context and payload in the callback. It also implements a callback for pongs.

schmidtw:
I didn't see that PR before.  Can #307 be merged?

nhooyr:
Not at the moment. I'm busy the next few weeks but I'll be working on this library again in January/February full time.

Issue #373, opened 2023-02-28
WASM: support DialOptions.HTTPClient


d-z-m:
Support configuring a custom `http.Client` for use in a Wasm deployment.

This would support making websocket connections through something like gvisor `netstack`.

An example of such an `http.Client` is [here](https://github.com/WireGuard/wireguard-go/blob/master/tun/netstack/examples/http_client.go).

d-z-m:
> For Wasm deployment, you cannot pass in a custom http.Client because we use the JS WebSocket API directly and it doesn't let you tunnel websocket connections through things like that.

I had thought you would simply use the `http.Client`'s dial functions to make the websocket connection.  Is that not the case?

d-z-m:
In this case, I don't need the browser API at all,  and would like it to behave just like it would in `dial.go` .

nhooyr:
> I had thought you would simply use the http.Client's dial functions to make the websocket connection. Is that not the case?

I can't make it any clearer.
Give this file a read https://github.com/nhooyr/websocket/blob/master/ws_js.go to understand what's going on.

d-z-m:
I understand that to make this library work in a Wasm environment, you hook into the JS WebSocket API directly. 

 The issue I'm dealing with is:  I would like to use the library to make WebSocket connections over a virtual network connection.  E.g. an `http.Client` with a custom `http.Transport`.

I understand that the golang Wasm `http.DefaultClient`  uses the Fetch API under the hood.  However, `http.Client` still supports custom `http.Transport`s when compiling to Wasm.

 Ideally, this library would just sit on top of that, and negotiate the websocket in the normal way it would when compiling for `linux x86` (i.e. invoking the `http.Client`'s `Dial` and `DialContext`, etc.

nhooyr:
Fair enough, let me look into it. It's been a few years since I wrote the Wasm code.

nhooyr:
Ah ok I understand your request now. I'll change things such that `*http.Client` is an accepted field but ignored with clear documentation if it doesn't have a custom transport.

nhooyr:
Or hold on actually so you're compiling not for browsers but for javascript in general. I'm not sure if I should support such a usecase. Then I need to add Accept support too.

nhooyr:
> I'm not sure exactly what you mean by this, can you elaborate? I am compiling for browsers, i.e. GOOS=js GOARCH=wasm

[...]

PR #449, opened 2024-05-04
fix: avoid importing net/http on wasm

paralin:
Using goweight: https://github.com/paralin/goweight

GOOS=js GOARCH=wasm goweight | less

Without this change:

  7.9 MB runtime
  6.6 MB net/http
  3.1 MB net
  3.0 MB crypto/tls
  2.0 MB reflect
  1.4 MB math/big
  1.3 MB crypto/x509
  935 kB os
  ...

With this change:

  7.9 MB runtime
  3.1 MB net
  2.0 MB reflect
  935 kB os
  ...

Slight modifications to avoid importing net/http reduce the binary size significantly.

### DIFF
--- a/ws_js.go
+++ b/ws_js.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"net"
-	"net/http"
 	"reflect"
 	"runtime"
 	"strings"
@@ -196,7 +195,7 @@ func (c *Conn) Ping(ctx context.Context) error {
 // Write writes a message of the given type to the connection.
 // Always non blocking.
 func (c *Conn) Write(ctx context.Context, typ MessageType, p []byte) error {
-	err := c.write(ctx, typ, p)
+	err := c.write(typ, p)
 	if err != nil {
 		// Have to ensure the WebSocket is closed after a write error
 		// to match the Go API. It can only error if the message type
@@ -210,7 +209,7 @@ func (c *Conn) Write(ctx context.Context, typ MessageType, p []byte) error {
 	return nil
 }
 
-func (c *Conn) write(ctx context.Context, typ MessageType, p []byte) error {
+func (c *Conn) write(typ MessageType, p []byte) error {
 	if c.isClosed() {
 		return net.ErrClosed
 	}
@@ -287,7 +286,7 @@ type DialOptions struct {
 // The passed context bounds the maximum time spent waiting for the connection to open.
 // The returned *http.Response is always nil or a mock. It's only in the signature
 // to match the core API.
-func Dial(ctx context.Context, url string, opts *DialOptions) (*Conn, *http.Response, error) {
+func Dial(ctx context.Context, url string, opts *DialOptions) (*Conn, *struct{}, error) {
 	c, resp, err := dial(ctx, url, opts)
 	if err != nil {
 		return nil, nil, fmt.Errorf("failed to WebSocket dial %q: %w", url, err)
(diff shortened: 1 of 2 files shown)


nhooyr (commented):
Hi, I understand the reason for this change but unfortunately I cannot merge it as it breaks the API. The API must be kept 1:1 between Go and Wasm and so Dial must return http.Response.

paralin:
Dial is a global function and the value returned on wasm is a dummy value.

Are people using these in interfaces? is that why? Other than that it seems like anyone using Dial would already be ignoring the dummy value on wasm.

nhooyr:
People could be wrapping this library in some kind of API/helper for their own code and they might be checking resp or passing it up the chain. This patch would break that.

I don't know if it's common but it doesn't seem far fetched to me. I don't have a better suggestion unfortunately. I'll leave this open for now, let's see if anyone else has thoughts.

mafredri:
Hey @paralin. Thanks for this proposal, the weight reduction is impressive but I believe this is an uphill battle. For instance, in #373 there's a request to introduce `*http.Client` as well as make the `DialOptions` struct the same for both wasm and non-wasm.

We also want to avoid breaking the API while we're v1, targeting API breakages for a potential v2 in the future is of course an option, but will not be done lightly.

With these constraints in mind, I don't see how we could support this. The only option I can think of is to let users opt-in via a build tag, say `GOOS=js GOARCH=wasm go build -tags slim`. As a result, we could radically change the wasm API and simplify the return values.

But before we consider adding more complexity to the code-base, I'd like to hear about use-cases where this weight reduction is necessary.

paralin:
In a web browser app using wasm with Go, any weight reduction is significant. 

While going through the list of packages taking up space in the binary, net/http was at the top of the list.

 6.6 MB net/http

> As of January 2024, the average download speed globally for mobile internet was 50 Mbps

That's around 1 second of time spent downloading net/http, and even a second delay loading a website feels like a lot.

Of course gzip / brotli compression reduces this significantly.

I understand the desire to keep API compatibility. I also use fetch on wasm to reduce binary size: https://github.com/aperturerobotics/util/blob/master/js/fetch/fetch.go

mafredri:
@paralin what you say makes sense but what I'm looking for is concrete use-cases. Code where you're using `websocket` and building for wasm, without importing anything from `net/http`.

If I'd write a program only using `websocket.Dial`, the difference wouldn't be all that big:

```
❯ ls -lha
total 7.1M
drwxr-xr-x 2 coder coder 4.0K Aug 21 12:24 ./
drwxr-xr-x 5 coder coder 4.0K Aug 21 12:24 ../
-rwxr-xr-x 1 coder coder 3.1M Aug 21 12:24 dial-master*
-rwxr-xr-x 1 coder coder 897K Aug 21 12:24 dial-master.gz*
-rwxr-xr-x 1 coder coder 2.4M Aug 21 12:24 dial-pr*
-rwxr-xr-x 1 coder coder 704K Aug 21 12:24 dial-pr.gz*
-rw-r--r-- 1 coder coder  271 Aug 21 12:23 main.go
```

Sure, `0.7M` is still significant, as is `193K` gzipped. The thing we need to evaluate though is if the cost of supporting this use-case is worth the gain. How big will the gain be? How easy is it to write and maintain a program with no imports from `http`, etc? And that's where real-world use-cases come in and would help a lot.

paralin:
I will go ahead and close as this is mega-stale :)

PR #452, opened 2024-06-04
Replace filepath.Match with path.Match

### LINKED ISSUE #451: Should use path.Match instead of filepath.Match
https://github.com/nhooyr/websocket/blob/v1.8.11/accept.go#L237

Typically, `path/filepath` is used for OS-specific paths/etc while `path` has a OS neutral `Match()` method that works exactly the same in most cases.

A network lib shouldn't be using methods which behavior might vary depending on the OS.

KianYang-Lee:
OS-specific behaviour is not necessary. This PR replaces filepath.Match with path.Match and also updated the documentation to reflect that.

Closes #451

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -14,7 +14,7 @@ import (
 	"net/http"
 	"net/textproto"
 	"net/url"
-	"path/filepath"
+	"path"
 	"strings"
 
 	"nhooyr.io/websocket/internal/errd"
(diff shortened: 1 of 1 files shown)


nhooyr:
@KianYang-Lee Nice this looks great thanks.

Could you rebase on `dev` to get rid of the extra commit?

KianYang-Lee:
@nhooyr rebased

KianYang-Lee:
@nhooyr is this good to merge?

mafredri (approved):
Looks great 🎉, thanks for the PR @KianYang-Lee!

PR #459, opened 2024-07-22
added detailed guide in echo readme.md example folder

sameer-gits:
I have just updated echo readme,nd example files because most go users are confused including please accept my pr, you can also edit it a little bit if you want or let me know if you need anything else, thanks for the lib

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,3 +1,3 @@
 module nhooyr.io/websocket
 
-go 1.19
+go 1.22.5
--- a/internal/examples/echo/README.md
+++ b/internal/examples/echo/README.md
@@ -19,3 +19,61 @@ and then reads all messages and writes them exactly as is back to the connection
 `server_test.go` contains a small unit test to verify it works correctly.
 
 `main.go` brings it all together so that you can run it and play around with it.
+
+## Detailed Guide
+
+### In main.go
+
+1. Project starts. `func main` as the entry point start the program.
+
+2. main func execute `run()` and `log.Fatal` is there is `err`
+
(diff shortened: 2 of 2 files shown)


sameer-gits:
why checks are failing i just added text in example docs..... i am new let me know if anyone sees my comment here

sameer-gits:
@nhooyr ...

go.mod
```diff
@@ -1,3 +1,3 @@
 module nhooyr.io/websocket
 
-go 1.19
+go 1.22.5
```
mafredri, line 3: We want to retain go 1.19 compatibility for now, bumping to 1.22 is too big a jump anyway as 1.20 would be the next one.

internal/examples/echo/README.md
```diff
@@ -19,3 +19,61 @@ and then reads all messages and writes them exactly as is back to the connection
...
 
 `main.go` brings it all together so that you can run it and play around with it.
+
+## Detailed Guide
+
+### In main.go
+
+1. Project starts. `func main` as the entry point start the program.
+
+2. main func execute `run()` and `log.Fatal` is there is `err`
```
mafredri, line 29: This is too much detail. This is obvious from reading the code.

mafredri (commented):
Hey, thank you for the contribution but I believe this is too much detail. If there's a specific problem let's address that but let's assume that basic knowledge of Go and code reading is a requirement.

mafredri:
Closing, feel free to open a new PR or issue if there's a specific concern you'd like addressed.

PR #472, opened 2024-08-15
accept: Add unwrapping for hijack like http.ResponseController

### LINKED ISSUE #455: Support (*http.ResponseController).Hijack()
Currently the library checks `http.ResponseWriter` for `http.Hijacker` support via [type assertion](https://github.com/nhooyr/websocket/blob/bd07a64be6809ef46b76fff5ef7263ef7641980f/accept.go#L108).

Is there a possibility we can use [`(*http.ResponseController).Hijack()`](https://pkg.go.dev/net/http#ResponseController.Hijack) instead? This way we can support `http.ResponseWriter` types that don't offer `http.Hijacker` interface but their underlying unwrapped writers do.

mafredri:
Since we rely on the connection not being hijacked too early (i.e. detecting the presence of http.Hijacker) to set headers, we must manually implement the unwrapping of the http.ResponseController. By doing so, we also retain Go 1.19 compatibility without build tags (still used in compat test, though).

Closes #455

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -105,7 +105,7 @@ func accept(w http.ResponseWriter, r *http.Request, opts *AcceptOptions) (_ *Con
 		}
 	}
 
-	hj, ok := w.(http.Hijacker)
+	hj, ok := hijacker(w)
 	if !ok {
 		err = errors.New("http.ResponseWriter does not implement http.Hijacker")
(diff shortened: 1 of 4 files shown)


code-asher (approved):
Sweet looks good to me!

PR #476, opened 2024-09-05
adds a lock on writeClose. closes #448

### LINKED ISSUE #448: failed: Close received after close
Hello, I recently encountered a problem. Can someone help me take a look？，
I used the [example](https://github.com/nhooyr/websocket/tree/master/internal/examples/chat) and successfully ran it, but I sent a ping to the server and received an error below:
```
WebSocket connection to 'ws://127.0.0.1:8888/subscribe' failed: Close received after close
```

FrauElster:


mafredri:
Hey, thanks for the PR. I do think we should take a slightly different approach, though. Instead of piggy-backing on `closeMutex` in `writeClose` and potentially introduce deadlocks due to multiple consumers of the mutex, we move the logic to `writeFrame` instead. Then we can rely on the already existing `writeFrameMu` and set a boolean for the close when `opCode == opClose`.

This has another nice benefit of catching the close message wherever it may originate from.

FrauElster:
Sure great idea. I just had this bug and need it fixed and in my experience fixing it and opening a PR is the fastest way. I did not read nor understand the whole codebase, just stepped through it with my debugger and implemented this quick and dirty solution. 

Just for clarification if you say we, do you mean you or me? Clearly  you already have the solution on your mind, but if you are busy I can also do it.

FrauElster:
As suggested I moved the writeSent read/write into writeFrame. 
However I felt bad to fail silently if a double Close occurred, and is not really an net.ErrClosed since the connection could still be open waiting for the close response, so I introduced a new ErrAlreadyClosed, but currently it is not handled anywhere.

conn.go
```diff
@@ -68,6 +68,7 @@ type Conn struct {
 	writeBuf       []byte
 	writeHeaderBuf [8]byte
 	writeHeader    header
+	closeSent      bool
```
mafredri, line 71: Review: I moved this here so that the it's clearer that `writeFrameMu` is what's protecting it.

write.go
```diff
@@ -249,6 +249,15 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, flate bool, opcode opco
...
 	defer c.writeFrameMu.unlock()
 
+	if c.closeSent {
+		select {
+		case <-c.closed:
+			return 0, net.ErrClosed
+		default:
+		}
+		return 0, errors.New("close sent")
+	}
```
mafredri, line 259: Review: I took the liberty to make a few changes here since the code needed to move around.

Any message after the close frame could be considered bad behavior and wasted traffic, so I'm thinking we just don't allow it throughout.

Also, since we don't have a precedent in this library for sentinel errors, we'll just move it inline for now. I don't think sentinel errors are bad per-se, and maybe we should introduce some in the future (but let's not do that in this PR).

Lastly, an important detail here is that we're _not_ letting this context potentially close the connection before the handshake is complete by doing the check early.

mafredri (commented):
Thanks for making the changes! I took the liberty to push a few myself to your branch, hope you don't mind. (See review comments.)

> Just for clarification if you say we, do you mean you or me? Clearly you already have the solution on your mind, but if you are busy I can also do it.

My thinking is that I'm happy to let you work on this since you started the PR, and hopefully land a contribution to the project. 

Side-note: I typically write/talk in _we_s as I see coding as a group effort, this way both successes and failures are shared and not the responsibility of any one individual. My hope is that it also sparks inclusivity by feeling like a group effort instead of one person making decisions.

---

TBH I think this PR is in pretty good shape now, but I think we should add at least one test testing this new behavior. Would you want to tackle that?

mitar:
Very interesting. I came to this PR because I am debugging our tests. In our tests we measure how many bytes we send over a websocket (using this library for both client and server in Go) and how many we get for a simple sending of text message `hi` from the client and then sending `hi` back from the server to the client.

With v1.8.10, before closing was rewritten, we had ``{"wsFromClient":16,"wsToClient":8}` bytes transmitted. With v1.8.12, we saw change to `{"wsFromClient":16,"wsToClient":12}` which I understood that the increase in `wsToClient` is because server is now properly closing. But then I noticed that it is not always like that and that sometimes it is `{"wsFromClient":24,"wsToClient":12}` because it seems client sends close twice. So I got to this PR and tried this PR and with this PR I am back to ``{"wsFromClient":16,"wsToClient":8}`, always. Not sure why not `{"wsFromClient":16,"wsToClient":12}`.

FrauElster:
Nice to hear that it makes things more predictable. :) 

@mafredri Yeah a test would be great. Sadly I am on a business trip until next monday, so until then this PR would be stall, unless you would not mind.

FrauElster:
So I had a moment to have look and wrote several approaches, but I have to admit that I find it kindof hard to test. I am not familiar enough with the code, so I'll describe some assumptions here.

My general idea is to open a server and a client connection, close one and count the amount of close messages received on both ends. Both are asserted to be one. 

Assumption 1:
The problem is that the exposed methods are highlevel and manage the closing handshake themself without returning it to the callee (e.g. `conn.Read()`) , so I cannot count them.

Assumption 2:
A connection ignores any messages received after the connection was closed. So it just gets discarded and I cannot count any more than 1 received closed message. (This behaviour differs e.g. compared to websocket client of chrome).

I am tempting to have a private callback on the connection options, but introducing private dependency injection just to be able to test it seems overkill, and introduces to much complexity for the benefits. (IMHO).

Second attempt would be to write a separat minimal websocket client / server that can count, but this seems like a bunch of work for such a simple test.

I saw that there are some helpers like EchoServer, one could also introduce something similar that can count opcodes, but also this would need a separat implementation since it is not in the websocket package. 

For now I am using my fork (thanks to [go mod replace](https://go.dev/ref/mod#go-mod-file-replace) thats pretty easy) to fix the problems i am currently experiencing in production. 

Looking forward to get some feedback on the testing issue.

FrauElster:
Another note:

your last commit makes my test (`go test -count=10 ./...`) gives me the "conn_test.go:623: failed to close WebSocket: failed to read frame header: EOF" again (sometimes). So the race conditions seems to be back in the current state.

mafredri:
> Another note:
> 
> your last commit makes my test (`go test -count=10 ./...`) gives me the "conn_test.go:623: failed to close WebSocket: failed to read frame header: EOF" again (sometimes). So the race conditions seems to be back in the current state.

Thanks @FrauElster, I can reproduce this. I took a look and I'm pretty sure that the fix simply surfaced a new issue. Calling `(*Conn).CloseRead(ctx)` seems to interfere with the close handshake, it hijacks the reader so that the close handshake isn't propagated during actual close. This in turn is closing the connection early so that `(*Conn).Close()` runs into `io.EOF`.

I appreciate you taking a look and thinking about ways to test for this, if you want though, I can take over as I want to look more closely at the way we close connections.

PS. Are you using `CloseRead(ctx)` yourself? And have you tried this branch as it is right now as a fix for the problems you're experiencing? If we disregard the failing test, have you been able to show that this fix doesn't work in your use-case?

FrauElster:
No I don't use CloseRead(ctx) in my application. I use the simple 

```golang
msgType, data, err := conn.Read(ctx)
if websocket.CloseStatus(err) != -1 { ... }
```

or 

```golang
conn.Close(websocket.StatusInternalError, string(reason))
```

mafredri:
@FrauElster I pushed a fix for the `failed to close WebSocket: failed to read frame header: EOF` issue. Let me know if you want me to open a separate PR since I'm pushing to your fork (I don't mind but don't want to cause conflicts for you).

@mitar I'm curious if c3613fc causes any change in the behavior you're seeing? I have noticed that there are still cases where the `writeTimeout` and `readTimeout` contexts aren't cleared and could potentially cause an abrupt closure of the connection. I'm not saying that's what's happening to you, just a potential cause.

FrauElster:
Sure go ahead. My production go.mod is pinned to a commit, so no worries. ;)

conn.go
```diff
@@ -69,13 +69,19 @@ type Conn struct {
...
+	// CloseRead state.
 	closeReadMu   sync.Mutex
 	closeReadCtx  context.Context
 	closeReadDone chan struct{}
 
-	closed  chan struct{}
+	closing atomic.Bool
 	closeMu sync.Mutex
-	closing bool
+	closed  chan struct{}
```
johnstcn, line 84: It's not immediately clear what `closeMu` is protecting here. I think it would be helpful to make this more explicit.

Ditto for `closing`, we could be more clear about what its purpose is here for future readers.

conn.go
```diff
@@ -69,13 +69,19 @@ type Conn struct {
...
+	closeReceivedErr error
+	closeSentErr     error
+
+	// CloseRead state.
 	closeReadMu   sync.Mutex
 	closeReadCtx  context.Context
 	closeReadDone chan struct{}
 
-	closed  chan struct{}
+	closing atomic.Bool
```
johnstcn, line 82: Would it be more idiomatic to have `closing chan struct{}` that we close to indicate that the connection has entered the "closing" state?

conn.go
mafredri: I can add a comment as it only protects the `closed` channel now. Previously it protected the `closing` boolean as well, but I changed it to atomic to avoid a deadlock in rare situations. For example when `msgWriter` is being closed at the same time as reading a close frame, resulting in writing a close frame as response. The mutexes between conn close and msg writers are somewhat intertwined and I'm thinking of doing a larger refactor to address this at some point.

conn.go
mafredri: I wouldn't say more idiomatic per-se, but at least here interacting with the channel would be racy, it would require protection via mutex which I wanted to avoid. See my other comment for a motivation for this.

read.go
```diff
@@ -217,57 +217,62 @@ func (c *Conn) readLoop(ctx context.Context) (header, error) {
...
 
-	select {
-	case <-c.closed:
-		return header{}, net.ErrClosed
-	case c.readTimeout <- context.Background():
+	c.closeStateMu.Lock()
+	closeReceivedErr := c.closeReceivedErr
+	c.closeStateMu.Unlock()
+	if closeReceivedErr != nil {
+		defer done()
```
mtojek, line 244: Shouldn't we make this behavior consistent - in this line you call `done()`, but at L248 you pass it up. Maybe we should refactor to pass it in both cases?

read.go
```diff
@@ -325,9 +330,19 @@ func (c *Conn) handleControl(ctx context.Context, h header) (err error) {
...
+
+	if !closeSent {
+		c.readMu.unlock()
+		_ = c.writeClose(ce.Code, ce.Reason)
+	}
+	if !c.casClosing() {
+		c.readMu.unlock()
+		_ = c.close()
+	}
 	return err
```
mtojek, line 346: consider a comment here explaining when `c.readMu.unlock()` is not expected to be called.

mtojek (approved):
Left a few non-blocking comments.

read.go
mafredri: It's not typical to return _both_ an error and a valid value (e.g. if there's an error caller assumes done is nil). The reason we defer done here is that we need to clean up the change we made (readTimeout ctx) if we're going to abort.

write.go
```diff
@@ -249,22 +248,36 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, flate bool, opcode opco
...
+			}
+			err = fmt.Errorf("failed to write frame: %w", err)
+		}
+	}()
+
+	c.closeStateMu.Lock()
+	closeSentErr := c.closeSentErr
+	c.closeStateMu.Unlock()
+	if closeSentErr != nil {
+		return 0, net.ErrClosed
```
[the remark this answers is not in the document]
mafredri: It's a trade-off between reproducible error and variable. For now this mimics behavior from before this PR so this should be fine.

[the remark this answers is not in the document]
mitar: We could return a joined error? `errors.Join`? Or wrap errors?

read.go
mtojek: I see, it just seems like a typical `try-finally` block, but in this case `finally` can be executed on different levels. I was wondering if we can make it consistent.

code-asher (approved):
It looks solid to me!

read.go
code-asher: I was thinking a similar thing, would it be weird not to return an error?  For the caller, seems fair to assume they must check `err` even if not returned since they are passing it in by reference?

idk if this is any better though, mostly just thinking out loud

```go
defer prepareRead(ctx, &err)()
if err != nil {
  return header{}, err
}
```

close.go
```diff
@@ -100,7 +100,7 @@ func CloseStatus(err error) StatusCode {
 func (c *Conn) Close(code StatusCode, reason string) (err error) {
```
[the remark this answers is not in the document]
mafredri: Good observations. Technically what you say is correct, however in practice multiple calls to Close should happen sequentially, so for all intents and purposes, the second should be no-op.

You do raise a good point that calling `Close` is not guaranteed to return a nil error every time, even if the connection has ended amicably. This has always been the case so I'm not sure we should do anything about it in this PR, but it's worth considering for a future refactor of closing behavior and what errors are returned and when.

read.go
mafredri: That's a clever suggestion @code-asher, I feel perhaps a bit too clever for me, though. Personally I like the explicitness of error handling in Go, albeit verbose, it always tells me when I need to handle an error. In the proposed syntax, I feel that err handling is not entirely obvious. I'll just go with what we have, as the only alternative I can think of is to undo the function and repeat the code between the two functions that use it, whichever is better feels like a coin toss. 😄

dmitry-novozhilov:
What about bump version?

PR #477, opened 2024-09-06
docs: Fix docs and examples related to r.Context() usage

### LINKED ISSUE #474: failed to close writer: context canceled
I'm using this library to create a zero-trust tunnel into a Kubernetes cluster, when I encountered this error message:

> failed to write msg: failed to close writer: failed to acquire lock: context canceled

So what happened was that when client closed the connection, I would send a "connection closed" message over the WebSocket connection, but I (mistakenly) used the request context for the write command.
This context would get cancelled somewhere around this Write call, sometimes after message was written but before it released the lock.

As a result, the lock would never get released and the websocket was no longer usable.

The solution was to create a new context for the write:
```go
	writeCtx, _ := context.WithTimeout(context.Background(), 2*time.Second)

	err = c.Websocket.Write(writeCtx, websocket.MessageText, marshal)
```

And I don't think I should have used the request context to begin with, but I do think it's not optimal behaviour that the websocket can be left in a broken state by mistimed context cancel.

In any case, just thought I'd share it. Thanks for maintaining this awesome library!

mafredri:
Contributes to #474

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ http.HandlerFunc(func (w http.ResponseWriter, r *http.Request) {
 	}
 	defer c.CloseNow()
 
-	ctx, cancel := context.WithTimeout(r.Context(), time.Second*10)
+	// Set the context as needed. Use of r.Context() is not recommended
+	// to avoid surprising behavior (see http.Hijacker).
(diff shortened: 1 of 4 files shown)


code-asher (approved):
Sorry for the delay!  It seems I did not set up my notifications for this repo correctly.

PR #481, opened 2024-09-09
chore: optimize ci

matifali:
Now, CI will only run on push to `master` and PRs to `master`.  Previously, it was running twice on each push to a PR branch, once for the PR to `master` and once for pushing to the PR branch.


mafredri (approved):
Thanks!

PR #487, opened 2024-09-09
hotfix: close received after close

fogrye:
waitCloseHandshake waits for client to respond with close frame but handleControl receives it and writes close again which may cause errors for some clients.

This is dump solution but it works in my case. I believe it can be done better, pls check it out.

Fixes: #448


fogrye:
I noticed more clever solution so ignore this one

PR #490, opened 2024-09-15
feat:(conn): add `RemoteAddr` and `LocalAddr` method

### LINKED ISSUE #489: Remote address
How we can get remote address of a connection?

kehiy:
- Fixes #489

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -11,12 +11,10 @@ import (
 	"fmt"
 	"io"
 	"log"
-	"net"
 	"net/http"
 	"net/textproto"
 	"net/url"
 	"path"
-	"strconv"
 	"strings"
 
 	"github.com/coder/websocket/internal/errd"
@@ -152,34 +150,12 @@ func accept(w http.ResponseWriter, r *http.Request, opts *AcceptOptions) (_ *Con
 	b, _ := brw.Reader.Peek(brw.Reader.Buffered())
 	brw.Reader.Reset(io.MultiReader(bytes.NewReader(b), netConn))
 
-	ipStr, portStr, err := net.SplitHostPort(r.RemoteAddr)
-	if err != nil {
-		err = errors.New("*http.Request contains invalid RemoteAddr")
-		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
-		return nil, err
-	}
-
-	ip := net.ParseIP(ipStr)
-	if ip == nil {
-		err = errors.New("*http.Request contains invalid IP address")
-		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
-		return nil, err
-	}
-
-	port, err := strconv.Atoi(portStr)
-	if err != nil {
-		err = errors.New("*http.Request contains invalid port number")
-		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
-		return nil, err
-	}
-
 	return newConn(connConfig{
 		subprotocol:    w.Header().Get("Sec-WebSocket-Protocol"),
 		rwc:            netConn,
 		client:         false,
 		copts:          copts,
 		flateThreshold: opts.CompressionThreshold,
-		remoteAddr:     &net.TCPAddr{IP: ip, Port: port},
 
 		br: brw.Reader,
 		bw: brw.Writer,
--- a/accept.go
+++ b/accept.go
@@ -156,6 +156,7 @@ func accept(w http.ResponseWriter, r *http.Request, opts *AcceptOptions) (_ *Con
 		client:         false,
 		copts:          copts,
 		flateThreshold: opts.CompressionThreshold,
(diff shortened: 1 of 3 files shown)


kehiy:
@mafredri hi, can you kindly review this, please?

conn.go
```diff
@@ -192,6 +195,11 @@ func (c *Conn) flate() bool {
 	return c.copts != nil
 }
 
+// RemoteAddr returns the remote address of websocket connection.
+func (c *Conn) RemoteAddr() string {
```
mafredri, line 199: I wonder if it would make sense to return a [`net.Addr`](https://pkg.go.dev/net#Addr) here, as it's a more well-known signature?

mafredri (commented):
Except for the suggestion inline, the PR looks good. Thanks.

I would be interested in knowing what your use-case is though, how is this feature beneficial to you?

kehiy:
@mafredri about use case, several programs need to collect IP address sometimes to use it for spam protection and this kind of stuff. sometimes we have to be aware of who we are talking too. or use it in logs to check whats going on better. in my case we are developing a nostr relay which can be aimed for spam attacks a lot.

conn.go
kehiy: if you think `net.Addr` is better, let me know so i can update it.

mafredri: The `RemoteAddr() string` signature is not in `net` or `x/net/websocket` AFAIK. If you look at the `x/net/websocket`, you can see that the return signature on Conn is `RemoteAddr() net.Addr`, not `string` as in this PR.

kehiy: ops, sorry. i thought it is its `.String()` method. yes, that's logical. let me update the pr.

kehiy:
@mafredri can you review again please?

accept.go
```diff
@@ -150,12 +152,34 @@ func accept(w http.ResponseWriter, r *http.Request, opts *AcceptOptions) (_ *Con
 	b, _ := brw.Reader.Peek(brw.Reader.Buffered())
 	brw.Reader.Reset(io.MultiReader(bytes.NewReader(b), netConn))
 
+	ipStr, portStr, err := net.SplitHostPort(r.RemoteAddr)
```
mafredri, line 155: A couple of notes:

- Parsing, and thus erroring here could potentially cause issues for current users (new behavior)
- We could call `netConn.RemoteAddr()` instead in our `RemoteAddr` func

accept.go
```diff
@@ -150,12 +152,34 @@ func accept(w http.ResponseWriter, r *http.Request, opts *AcceptOptions) (_ *Con
...
+	if err != nil {
+		err = errors.New("*http.Request contains invalid RemoteAddr")
+		http.Error(w, http.StatusText(http.StatusBadRequest), http.StatusBadRequest)
```
mafredri, line 158: Note: Since the connection has been hijacked, we can't use `http.Error` or `w` here. If we'd want to tell the client about this issue, it should happen before switching protocols and hijacking.

(That said, this new behavior shouldn't return any errors here, see previous notes.)

conn.go
```diff
@@ -192,6 +195,11 @@ func (c *Conn) flate() bool {
 	return c.copts != nil
 }
 
+// RemoteAddr returns the remote address of websocket connection.
+func (c *Conn) RemoteAddr() net.Addr {
+	return c.remoteAddr
```
mafredri, line 200: This should work for both client and server connections, would be good to add tests for both.

mafredri:
Since we're adding `RemoteAddr() net.Addr`, I think it would make sense to add `LocalAddr() net.Addr` too, do you want to include that in this PR?

accept.go
kehiy: got it.

conn.go
kehiy: okkk

kehiy:
@mafredri so, based on your review, i found that we have a `netConn` as well which i didn't saw it before. so this one is good. to achieve what you say in reviews we have 3 ways:

1. keeping `netConn` in `Conn` struct.
2. keeping `RemoteAddr` and/or `LocalAddr` in `Conn` struct.
3. change `rwc` in `Conn` to keep the `netConn`.

which one makes sense in our case?

kehiy:
> Since we're adding `RemoteAddr() net.Addr`, I think it would make sense to add `LocalAddr() net.Addr` too, do you want to include that in this PR?

i would include it in this pr.

mafredri:
You can just do a similar check as done here: https://github.com/coder/websocket/blob/6c8e3ab3985f154e911339dba08291ff0343b492/netconn_notjs.go#L8-L20

While you're at it, the linked code can be changed to just return `nc.c.RemoteAddr()`, etc.

We should also add these methods to the WASM implementation too, even if we can't return any useful information.

kehiy:
@mafredri now both are added based on what you suggest. adding to wasm implementation is remained correct? should it be included in this pr? i need to look at wasm implementation to find out how can i do that.

PR #499, opened 2024-11-07
chore(.github): group dependabot PRs and reduce frequency for thirdpary

mafredri:
This project has zero dependencies, so dependabot PRs are mainly noise. This PR ensures all thirdparty deps are grouped together and changes the frequency from weekly to monthly.


.github/dependabot.yml
```diff
@@ -1,9 +1,38 @@
 version: 2
 updates:
-  - package-ecosystem: 'gomod'
+  # Track in case we ever add dependencies.
+  - package-ecosystem: "gomod"
     directories:
```
mtojek, line 5: could that be just `directory: "/"`?

.github/dependabot.yml
```diff
@@ -1,9 +1,38 @@
...
+      - "/"
     schedule:
-      interval: 'weekly'
+      interval: "weekly"
+    commit-message:
+      prefix: "chore"
+
+  # Keep examples up-to-date.
+  - package-ecosystem: "gomod"
+    directories:
```
mtojek, line 14: nit: same question, `directory:`

mtojek (approved):
minor comment, otherwise 👍

matifali (approved):
Why can't we group while keeping the single config?
I guess GitHub doesn't support that.

mafredri:
> Why can't we group while keeping the single config? I guess GitHub doesn't support that.

I kept the groups separate to signify importance, where importance goes in the order of `/ > /internal/examples > /internal/thirdparty`. We don't have deps in `/` though, so that's just future-proofing.

Even `/internal/examples` is just one dep, so that's future proofing as well.

If there's opposition to this then we can remove `/` entirely and just use a single one:

```
  - package-ecosystem: 'gomod'
    directories:
      - '/internal/examples'
      - '/internal/thirdparty'
    schedule:
      interval: 'monthly'
    commit-message:
      prefix: 'chore'
    labels: []
    groups:
      internal:
        patterns:
          - '*'
```

Would that be preferable @matifali @mtojek ?

PR #501, opened 2024-11-13
abolish an extra goroutine

### LINKED ISSUE #411: Use context.AfterFunc to remove extra goroutine per connection used for context cancellation
See https://github.com/golang/go/issues/57928

Wonderful. Now there can be no performance disadvantage to my library vs gorilla.

See also 8060f3a

andriibeee:
closes #411

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -179,7 +179,7 @@ func (c *Conn) clearWriteTimeout() {
 
 func (c *Conn) setupReadTimeout(ctx context.Context) {
 	hammerTime := context.AfterFunc(ctx, func() {
-		defer c.close()
+		c.close()
 	})
 
 	if closer := c.readTimeoutCloser.Swap(hammerTime); closer != nil {
--- a/conn.go
+++ b/conn.go
@@ -52,9 +52,8 @@ type Conn struct {
 	br             *bufio.Reader
 	bw             *bufio.Writer
 
-	readTimeout     chan context.Context
-	writeTimeout    chan context.Context
-	timeoutLoopDone chan struct{}
+	readTimeoutCloser  atomic.Value
+	writeTimeoutCloser atomic.Value
 
 	// Read state.
 	readMu         *mu
@@ -104,10 +103,6 @@ func newConn(cfg connConfig) *Conn {
 		br: cfg.br,
 		bw: cfg.bw,
 
-		readTimeout:     make(chan context.Context),
-		writeTimeout:    make(chan context.Context),
-		timeoutLoopDone: make(chan struct{}),
-
 		closed:      make(chan struct{}),
 		activePings: make(map[string]chan<- struct{}),
 	}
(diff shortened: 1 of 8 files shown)


andriibeee:
sorry for all the commit spam, i've read a tweet sometime ago that github actions is a fiend to clean git commit history, while i think it's true it's my fault i'm too eepy and couldn't be bothered to run the workflow locally, check up .mod files and not to forgor about running go mod tidy. think this is like your average dark souls bonfire-to-"you died" message cycle but somewhat upside down, fix issue -> commit -> push -> wait for ci to fail -> oh my gosh here we go again.

please review my pr carefully, i am super new to the codebase although i've been using this library for a while and i might've messed up choosing atomic.Value vs atomic.Pointer but the latter looks more messy to me code wise and i don't think there is any sense to keep a pointer to function instead of keeping it straight as a value, maybe someone smarter than me and/or more acquainted with the codebase than me could propose anything. i tried my best to do the best choices and keep the style consistent. maybe you'll find an nit or two but i'm up for fixing them up.

conn.go
```diff
@@ -52,9 +52,8 @@ type Conn struct {
 	br             *bufio.Reader
 	bw             *bufio.Writer
 
-	readTimeout     chan context.Context
-	writeTimeout    chan context.Context
-	timeoutLoopDone chan struct{}
+	readTimeoutCloser  atomic.Value
+	writeTimeoutCloser atomic.Value
```
ash2k, line 56: I think `atomic.Pointer` would be cleaner. `atomic.Value` still stores the value as a pointer, it's just not explicit. The upside of `atomic.Pointer` is that it requires no type casting.

conn.go
```diff
@@ -164,26 +157,42 @@ func (c *Conn) close() error {
...
-		case readCtx = <-c.readTimeout:
-
-		case <-readCtx.Done():
-			c.close()
-			return
-		case <-writeCtx.Done():
-			c.close()
-			return
+func (c *Conn) clearWriteTimeout() {
+	if closer := c.writeTimeoutCloser.Load(); closer != nil {
```
ash2k, line 173: Shouldn't this be a `swap` too to remove the reference to the function to let GC take it? Same below.

conn.go
```diff
@@ -164,26 +157,42 @@ func (c *Conn) close() error {
 	return err
 }
 
-func (c *Conn) timeoutLoop() {
-	defer close(c.timeoutLoopDone)
+func (c *Conn) setupWriteTimeout(ctx context.Context) {
+	hammerTime := context.AfterFunc(ctx, func() {
+		c.close()
```
ash2k, line 162: I wonder if it'd be possible for this function to also remove itself from `c.writeTimeoutCloser`? Perhaps it can put a `nil` in it / or swap it with a `nil` and run what was in it if it's not `nil`. That way we'll let GC collect the function once its ran. Same below.

Well, it just needs to call `clearWriteTimeout()`.

Jacalz:
Any update on this @andriibeee?

mafredri:
Thanks for working on this andriibeee! Your work inspired #532 which just got merged. 🎉

PR #508, opened 2025-01-21
Fix build with Go 1.24

flyn-org:
Go 1.24 provides stricter checking that forbids passing a variable as a format string to Printf-like functions with no other arguments. Remove instances of this. See also:

https://tip.golang.org/doc/go1.24#vet

### DIFF
--- a/autobahn_test.go
+++ b/autobahn_test.go
@@ -92,7 +92,7 @@ func TestAutobahn(t *testing.T) {
 		}
 	})
 
-	c, _, err := websocket.Dial(ctx, fmt.Sprintf(wstestURL+"/updateReports?agent=main"), nil)
+	c, _, err := websocket.Dial(ctx, wstestURL+"/updateReports?agent=main", nil)
 	assert.Success(t, err)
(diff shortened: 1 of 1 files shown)


autobahn_test.go
```diff
@@ -92,7 +92,7 @@ func TestAutobahn(t *testing.T) {
 		}
 	})
 
-	c, _, err := websocket.Dial(ctx, fmt.Sprintf(wstestURL+"/updateReports?agent=main"), nil)
+	c, _, err := websocket.Dial(ctx, fmt.Sprint(wstestURL+"/updateReports?agent=main"), nil)
```
igolaizola, line 95: I would use just `wstestURLL+"/updateReports?agent=main"` or `fmt.Sprintf("%s/updateReports?agent=main", wstestURL)`. `fmt.Sprint` usage seems redundant here.

flyn-org: Good point. I just force pushed a revision.

mafredri (approved):
Thanks for the fix @flyn-org!

Issue #512, opened 2025-01-29
CI failing when launching TestWasm with wasmbrowsertest


igolaizola:
The CI is currently failing with this error when trying to launch TestWasm with wasmbrowsertest:

```
No usable sandbox! If you are running on Ubuntu 23.10+ or another Linux distro that has disabled unprivileged user namespaces with AppArmor, see https://chromium.googlesource.com/chromium/src/+/main/docs/security/apparmor-userns-restrictions.md. Otherwise see https://chromium.googlesource.com/chromium/src/+/main/docs/linux/suid_sandbox_development.md for more information on developing with the (older) SUID sandbox. If you want to live dangerously and need an immediate workaround, you can try using --no-sandbox.
```

I have 2 proposals to fix this:

## Disable AppArmor

Check this PR: https://github.com/coder/websocket/pull/511

This looks like the way to go. There is an open PR with the same approach here: https://github.com/agnivade/wasmbrowsertest/pull/72

## Add `--no-sandbox`
This requires that this PR is first merged to wasmbrowsertest https://github.com/agnivade/wasmbrowsertest/pull/73
Meanwhile, I have created a PR https://github.com/coder/websocket/pull/510 that points to my fork of wasmbrowsertest to confirm that it works.

PR #513, opened 2025-01-29
ci: disable AppArmor on daily and static workflows

igolaizola:
AppArmor was disabled in CI workflows, but it was not disabled in the daily and static workflows.

References #512


igolaizola:
Build in master is failing, this should fix it.
cc @mafredri

mafredri (approved):
Good catch, thanks for fixing these too!

PR #515, opened 2025-01-30
feat: implement websocket.Accept in js (#373)

shynome:
https://github.com/coder/websocket/issues/373#issuecomment-1456574498

DialOptions.HTTPClient also supported


mafredri:
@shynome thanks for the PR. There's quite a lot going on here, could you explain in broad terms what changes you've made, and why. And what new use-cases do these changes enable?

I have some reservations about making such a large API change and introducing new exported types.

shynome:
yes, I'm buillding a vpn based on webrtc, it can be running in browser.

the srouce branch: https://github.com/shynome/websocket/commits/feature/js-accept/

I have do the below things

1. rename `Conn` to `StdConn` for split `BrowserConn` and `GoConn`
2. Add the `type Conn interface`, but change `type Conn sturct` to `type Conn interface` is make api compatible broken,
   so I rename `type Conn interface` to `type Stream interface` (maybe other name is better clearly, could you provide one?), and use `the Conn struct` wrap `type Stream interface`
3. now we can export `websocket.Accept` in js, because now we have `StdConn` in js

shynome:
I know this PR has a lot of changes, so it might not be accepted, but in reality, the changes I made are quite few.

First, add the type Conn(Stream) interface to unify StdConn and BrowserConn. Then, export StdConn and BrowserConn so that users know which actual Conn they are working with.

Finally, thanks for your review!

PR #532, opened 2025-08-01
Use `context.AfterFunc()` instead of a goroutine to track timeouts

ash2k:
This is a reworked version of https://github.com/coder/websocket/pull/501 with some minor improvements.

Closes https://github.com/coder/websocket/issues/411

Closes https://github.com/coder/websocket/pull/501

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -51,9 +51,8 @@ type Conn struct {
 	br             *bufio.Reader
 	bw             *bufio.Writer
 
-	readTimeout     chan context.Context
-	writeTimeout    chan context.Context
-	timeoutLoopDone chan struct{}
+	readTimeoutStop  atomic.Pointer[func() bool]
+	writeTimeoutStop atomic.Pointer[func() bool]
 
 	// Read state.
 	readMu         *mu
@@ -113,10 +112,6 @@ func newConn(cfg connConfig) *Conn {
 		br: cfg.br,
 		bw: cfg.bw,
 
-		readTimeout:     make(chan context.Context),
-		writeTimeout:    make(chan context.Context),
-		timeoutLoopDone: make(chan struct{}),
-
 		closed:         make(chan struct{}),
 		activePings:    make(map[string]chan<- struct{}),
 		onPingReceived: cfg.onPingReceived,
@@ -143,8 +138,6 @@ func newConn(cfg connConfig) *Conn {
 	runtime.SetFinalizer(c, func(c *Conn) {
 		c.close()
 	})
-
-	go c.timeoutLoop()
 
 	return c
 }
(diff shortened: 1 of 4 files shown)


ash2k:
PTAL @mafredri @Jacalz

Jacalz (approved):
Code looks good and the changes work well in my testing. Thanks :+1:

conn.go
```diff
@@ -175,27 +168,34 @@ func (c *Conn) close() error {
...
 }
 
-func (c *Conn) timeoutLoop() {
-	defer close(c.timeoutLoopDone)
+func (c *Conn) setupWriteTimeout(ctx context.Context) {
+	stop := context.AfterFunc(ctx, func() {
+		c.clearWriteTimeout()
+		c.close()
+	})
+	swapTimeoutStop(&c.writeTimeoutStop, &stop)
```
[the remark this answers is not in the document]
ash2k: Hm, I do not understand the scenario you are describing. When we are replacing a context, we call `stop()` that we are removing - line 198 (`(*oldStop)()`). Could you describe the scenario that you have in mind step by step? Thank you for the review!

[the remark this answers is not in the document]
mafredri: Oh, sorry, looks like I made a short-circuit in my thinking. You're absolutely right, there's no issue there 👍🏻.

read.go
```diff
@@ -220,22 +220,24 @@ func (c *Conn) readLoop(ctx context.Context) (header, error) {
...
 
 	done := func() {
 		select {
 		case <-c.closed:
 			if *err != nil {
 				*err = net.ErrClosed
 			}
-		case c.readTimeout <- context.Background():
+		default:
+			c.clearReadTimeout()
```
mafredri, line 240: This should be moved outside the select, if the connection is closed, we don't want to leave the timeout behind.

write.go
```diff
@@ -271,12 +271,14 @@ func (c *Conn) writeFrame(ctx context.Context, fin bool, flate bool, opcode opco
...
-	case c.writeTimeout <- ctx:
+	default:
 	}
+	c.setupWriteTimeout(ctx)
 	defer func() {
 		select {
 		case <-c.closed:
-		case c.writeTimeout <- context.Background():
+		default:
+			c.clearWriteTimeout()
```
mafredri, line 281: Same here, move outside the select.

read.go
ash2k: Yep, that was a bug! Thanks for catching. Pushed the fix.

mafredri (approved):
Looks good, thanks again!

ash2k:
Thanks for merging! If you could tag a release, I'd deploy to prod and see what happens 🤞

mafredri:
@ash2k Released as [`v1.8.14`](https://github.com/coder/websocket/releases/v1.8.14)!

Issue #220, opened 2020-04-02
klauspost/compress is 21 megabytes


nhooyr:
Some user's reported an issue with the dependency being large https://github.com/diamondburned/arikawa/issues/11

I confirmed, the dependency is 21 megabytes.

I think the best approach would be to vendor only the code we need from klauspost/compress.

cc @klauspost

diamondburned:
Seeing how the `compress` PR used [github.com/klauspost/compress/flate](https://godoc.org/github.com/klauspost/compress/flate), can the stdlib [compress/flate](https://golang.org/pkg/compress/flate/) be used instead?

nhooyr:
stdlib compress/flate is significantly slower. See https://github.com/klauspost/compress/pull/216#issuecomment-586638251

nhooyr:
And allocates much more memory.

diamondburned:
I think 2x slower isn't that big of a deal. If anything, the library can also allow an interface to give developers choices when it comes to performance-intensive tasks.

diamondburned:
Alternatively, I think @klauspost should really split up his repositories. Seeing how the largest file seems to be from `compress/zstd`, he could probably split 5 of his subpackages into 5 repositories.

![image](https://user-images.githubusercontent.com/8463786/78310908-2b009980-7504-11ea-90b8-ec314907194c.png)

pollen5:
Vendoring only needed code sounds good.

klauspost:
No. If 21 megs is a problem for you, do what you want.

nhooyr:
> The question is why is he pushing test data to the repo.

Everyone pushes tests to repos.

Like @klauspost said, I'm not entirely sure this is a problem worth solving.

nhooyr:
I think the default should use stdlib and then I can maintain a separate branch if there's enough interest until/if stateless compression ends up in the stdlib. It's always nice to have zero dependencies as well.

PR #534, opened 2025-08-12
Support client_max_window_bits and server_max_window_bits compression options

### LINKED ISSUE #443: unsupported permessage-deflate parameter: "client_max_window_bits=15" from client
If a websocket server reply with extension `permessage-deflate; client_max_window_bits=15` an error happens:
```
unsupported permessage-deflate parameter: "client_max_window_bits=15"
```
at
https://github.com/nhooyr/websocket/blob/bd07a64be6809ef46b76fff5ef7263ef7641980f/dial.go#L268-L300

hexdigest:
This is related and fixes #443

### DIFF
--- a/accept.go
+++ b/accept.go
@@ -15,6 +15,7 @@ import (
 	"net/textproto"
 	"net/url"
 	"path"
+	"strconv"
 	"strings"
 
 	"github.com/coder/websocket/internal/errd"
@@ -298,15 +299,34 @@ func acceptDeflate(ext websocketExtension, mode CompressionMode) (*compressionOp
 		case "server_no_context_takeover":
 			copts.serverNoContextTakeover = true
 			continue
-		case "client_max_window_bits",
-			"server_max_window_bits=15":
+		case "client_max_window_bits":
+			copts.clientMaxWindowBits = 15 // default
+			continue
+		case "server_max_window_bits":
+			copts.serverMaxWindowBits = 15 // default
 			continue
 		}
 
 		if strings.HasPrefix(p, "client_max_window_bits=") {
-			// We can't adjust the deflate window, but decoding with a larger window is acceptable.
+			// We don't need to change decoder settings; larger window decoder can read smaller windows.
+			if v, err := strconv.Atoi(strings.TrimPrefix(p, "client_max_window_bits=")); err == nil {
+				if v >= 8 && v <= 15 {
+					copts.clientMaxWindowBits = v
+				}
+			}
 			continue
 		}
+
+		if strings.HasPrefix(p, "server_max_window_bits=") {
+			vstr := strings.TrimPrefix(p, "server_max_window_bits=")
+			v, err := strconv.Atoi(vstr)
+			if err != nil || v < 8 || v > 15 {
+				return nil, false // invalid per RFC
+			}
+			copts.serverMaxWindowBits = v
+			continue
+		}
+
 		return nil, false
 	}
 	return copts, true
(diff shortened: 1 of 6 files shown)


hexdigest:
Hey guys these :point_up:  are failing due to thirdparty dependency.

@nhooyr [the third party dependency](https://github.com/klauspost/compress/) I introduce here is a well known drop in replacement for compress/flate. It also has constructors that allows you to specify the size of the window.

mafredri:
@hexdigest is this something you still want to work on?

hexdigest:
> @hexdigest is this something you still want to work on?

Yep, I can do it. My only concern about the options approach is that depending on the provided flate implementation we should enable or disable certain compression options. But I guess this could be a part of flate abstraction interface (that can return a list of supported options) and manage pools.

Does it make sense to you?

mafredri:
What kind of use-cases are you considering?

Off the top of my head I could see us having something like this:

```go
// Match stdlib flate.Writer, add release.
type FlateWriter interface {
	io.WriteCloser
	Flush() error
	Reset(io.Writer)
	Release() // Allow backing implementation to restore to pool.
}

type Options struct {
	FlateWriter func(w io.Writer, bits int) FlateWriter
}
```

But open to other suggestions, especially if this doesn't cover some use-case.

hexdigest:
> What kind of use-cases are you considering?
> 
> Off the top of my head I could see us having something like this:
> 
> ```go
> // Match stdlib flate.Writer, add release.
> type FlateWriter interface {
> 	io.WriteCloser
> 	Flush() error
> 	Reset(io.Writer)
> 	Release() // Allow backing implementation to restore to pool.
> }
> 
> type Options struct {
> 	FlateWriter func(w io.Writer, bits int) FlateWriter
> }
> ```

I was thinking about something like this :point_up: maybe with the one minor difference which is putting Release() logic into Close() method of WriteCloser.

But the thing that I was talking about in the above message is that depending on the FlateWriter implementation the websocket implementation (e.g. *_max_window_bits) might or might not accept certain compression options passed by the client. So maybe we can add a method to this FlateWriter interface that returns compression options that it supports.

WDYT?

mafredri:
> I was thinking about something like this ☝️ maybe with the one minor difference which is putting Release() logic into Close() method of WriteCloser.

Yeah, that's honestly better 👍🏻

> But the thing that I was talking about in the above message is that depending on the FlateWriter implementation the websocket implementation (e.g. *_max_window_bits) might or might not accept certain compression options passed by the client. So maybe we can add a method to this FlateWriter interface that returns compression options that it supports.
> 
> WDYT?

That could overcomplicate the interface. If I'm understanding you correctly, I think we could cover the use-case by adding `CompressionServerWindowMaxBits` and `CompressionServerWindowMinBits` to `AcceptOptions`?

Usage would then look something like:

```go
c, err := websocket.Accept(w, r, &websocket.AcceptOptions{
	CompressionServerWindowMinBits: 9,
	CompressionServerWindowMaxBits: 15,
	FlateWriter:                    kpFlateWriter(kpflate.DefaultCompression),
})

func kpFlateWriter(level int) func(io.Writer, int) FlateWriter {
    return func(dst io.Writer, bits int) FlateWriter {
        // ...
    }
}
```

Does that make sense?

mafredri:
FYI @hexdigest I came across today that this library used to integrate `klauspost/compress`, removed in #240. The reason seems to have been that it was too big (#220). I see this as validating for the approach we've discussed here.

PR #535, opened 2025-08-15
refactor: replace fmt.Errorf with MessageTooBigError

DanielleMaywood:


read.go
```diff
@@ -17,6 +17,14 @@ import (
 	"github.com/coder/websocket/internal/util"
 )
 
+type MessageTooBigError struct {
+	Limit int64
+}
+
+func (e MessageTooBigError) Error() string {
+	return fmt.Sprintf("read limited at %v bytes", e.Limit)
```
johnstcn, line 25: I want to change this to `%d`, but it was `%v` before so let's maybe leave it that way 😂

johnstcn (commented):
Deferring approval to @mafredri

read.go
DanielleMaywood: Better safe than sorry 😅

read.go
```diff
@@ -17,6 +17,14 @@ import (
 	"github.com/coder/websocket/internal/util"
 )
 
+type MessageTooBigError struct {
```
mafredri, line 20: If we need this, I'd still like to avoid introducing new exported error types. I think a sentinel error would be better here as they're a bit easier to use as well via `errors.Is`.

```go
var ErrMessageTooBig = errors.New("websocket: message too big")
```

That also doesn't stop us from adding context (limit) via:

```go
err = fmt.Errorf("%w: %w", ErrMessageTooBig, context)
```

read.go
```diff
@@ -520,7 +528,7 @@ func (lr *limitReader) Read(p []byte) (int, error) {
 	}
 
 	if lr.n == 0 {
-		err := fmt.Errorf("read limited at %v bytes", lr.limit.Load())
+		err := MessageTooBigError{Limit: lr.limit.Load()}
```
mafredri, line 531: Could you also update the code-path in `ws_js.go`?

mafredri (commented):
Considering this code-path closes the connection, what use-case are we addressing by detecting the specific error?

read.go
mafredri: I understand your concern. However, string matching on errors is never recommended and consumers should expect it to be brittle, especially when it contains a number that may change. That said, we could retain the error format meaning if consumers use "containers" rather than "==" they would be fine:

```go
fmt.Errorf("%w: read limited at %d bytes", ErrMessageTooBig, c.msgReadLimit.Load())
```

I've been giving this some more thought. And I really dislike introducing YAAS (yet another API surface?).
- We already have `StatusMessageTooBig`
- We already close the connection
- We call `c.writeError` which calls `c.writeClose` _and_ returns an error.

Why don't we just do:

```go
msg := fmt.Sprintf("read limited at %d bytes", c.msgReadLimit.Load())
err := c.writeClose(StatusMessageTooBig, msg)
return err
```

This will return a `CloseError`, but that is actually more correct IMO. The status of `CloseError` can be checked for `StatusMessageTooBig`. Thoughts?

Calling `.Error()` on the `CloseError` will return `status = 1009 and reason = read limited at N bytes`. This is good enough for a string contains but if we _really_ need to do 100% backwards compatibility. We can create a private error type that stringifies the backwards compatible error, but implements custom methods for error is/as to match on `CloseError`. I don't think we need to go that far though.

mafredri: Actually looks like I skimmed `writeClose`, it doesn't return the `CloseError`, but constructing it would be an option.

But maybe returning `CloseError` from read would be too unexpected, so sentinel approach is perhaps best after all.

Issue #528, opened 2025-04-18
Any chance of bubbling up HTTP errors?


anderspitman:
When `websocket.Dial` fails, my app wants to take differents actions depending on the HTTP status code, and ideally the body too. Is there any chance of modifying Dial to return this information, or perhaps adding a new function which does so?

PR #548, opened 2025-12-15
feat: introduces an option for a custom body size in websocket.Dial function

IgorSilvestre:
#### Why

Feature asked by this issue:
- https://github.com/coder/websocket/issues/528

 #### What

  - Introduce `DialOptions.MaxErrorResponseBodyBytes` so callers can opt into more (or zero) error-body
    bytes when the handshake fails, enabling better branching on status/body while keeping the default
    1024-byte limit.
  - Preserve existing success-path behavior and safety: response bodies stay nil on success; 3s timeout
    still guards error-body reads; disabled option closes without reading.

  #### Key Changes

  - dial.go: use new option to bound error-body capture; skip capture if set < 0; guard nil bodies;
    still close original body with 3s fail-safe.
  - dial_test.go: new/updated tests covering default cap, custom cap, disabled capture (ensures close),
    status code presence, and nil-body safety.

#### Testing
I did some testing with a dummy server responding a fixed status 400 and a 1mb body:
image

image

### DIFF
--- a/dial.go
+++ b/dial.go
@@ -160,7 +160,7 @@ func dial(ctx context.Context, urls string, opts *DialOptions, rand io.Reader) (
 	respBody := resp.Body
 	resp.Body = nil
 	defer func() {
-		if err != nil {
+		if err != nil && respBody != nil {
 			// Capture a limited portion of the response body for easier debugging,
 			// following the limit configured by MaxErrorResponseBodyBytes.
 			limit := opts.MaxErrorResponseBodyBytes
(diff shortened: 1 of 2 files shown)


IgorSilvestre:
> Closing this as it does not address #528, which does not request an increase in body size.

Fair, I assumed in 
> and ideally the body too. Is there any chance of modifying Dial to return this information"

that he would want to return more of the body, seeing its already returning 1024b.

IgorSilvestre:
> @IgorSilvestre no worries. Thanks for attempting a fix. We are very conservative with expanding API surface in this project and there would have to be a real and obvious need if we were to add something.

On reading other issues I saw that minimizing the API is something important for you, no problem.

I really like this project, if ever need a hand with something just ask.

PR #549, opened 2025-12-16
ci: remove dev branch jobs and add static.sh

mafredri:
Remove bench-dev and test-dev jobs from daily workflow since the dev
branch no longer exists.

Add ci/static.sh script to generate coverage badge for GitHub Pages.

Refs #539

### DIFF
--- a/ci/static.sh
+++ b/ci/static.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -eu
+
+cd -- "$(dirname "$0")/.."
+
+if [[ ! -f ./ci/out/profile.txt ]]; then
+	echo "No coverage profile found, run 'make test' first"
+	exit 1
+fi
+
+if [[ ! -f ./ci/out/coverage.html ]]; then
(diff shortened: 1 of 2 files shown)


ci/static.sh
```diff
@@ -0,0 +1,20 @@
...
+
+if [[ ! -f ./ci/out/coverage.html ]]; then
+	echo "No coverage report found, run 'make test' first"
+	exit 1
+fi
+
+rm -rf ./ci/out/static
+mkdir -p ./ci/out/static
+cp ./ci/out/coverage.html ./ci/out/static/coverage.html
+percent=$(go tool cover -func ./ci/out/profile.txt | tail -n1 | awk '{print $3}' | tr -d '%')
```
mtojek: set pipefail?

ci/static.sh
```diff
@@ -0,0 +1,20 @@
...
+wget -O ./ci/out/static/coverage.svg "https://img.shields.io/badge/coverage-${percent}%25-success"
```
mtojek: isn't curl more universal option?

PR #550, opened 2025-12-16
fix(dial): verify HTTP/1.x protocol in verifyServerResponseH1

mafredri:
Add resp.ProtoMajor check in verifyServerResponseH1 for consistency
with verifyServerResponseH2 which already performs this check.

Refs #539


code-asher (approved):
Sweet!  The http prefix makes a lot of sense too for consistency.

Issue #435, opened 2024-02-21
Tiny frames when compression is enabled


andersroos:
Hi I am experiencing a pretty weird behavior and I am unsure if this is intentional or not.

We have had problems with bad client performance when reading data from a websocket in a game client based on Unreal Engine.

I have investigated this and with compression enabled each message is chopped up in tiny frames (the problem surfaced when we run with 1.8.7 which had compression on by default). The problem with Unreal is that it only reads one frame per client tick (~8 milliseconds) by default.

When I inspect this in Wireshark it looks like every other frame has a 236 byte payload and every other has a 4 byte payload.

I also run it in the debugger and it looks like the server will send a frame for each write call from the flate lib (github.com/klauspost/compress) and it has a hard coded `bufferFlushSize` set to 240 which in turn causes all those individual calls to `writeFrame`.

Is this intentional?

nhooyr:
Are you using `Write` or `Writer`? `Write` should write in a single frame whereas Writer unfortunately does write in two frames.

If you are using `Write` it's possible I forgot to add the code necessary to make it use a single frame with compression enabled.

andersroos:
Thanks for the response.

I have tested on 1.8.10 and it is the same behavior if compression is enabled (`CompressionMode: websocket.CompressionNoContextTakeover`). 

The code uses `Write` (this method: https://github.com/nhooyr/websocket/blob/e3a2d32f704fb06c439e56d2a85334de04b50d32/write.go#L44).

I can make a minimal example to reproduce if it's helpful.

nhooyr:
No need I see it, the problem is here: https://github.com/nhooyr/websocket/blob/e3a2d32f704fb06c439e56d2a85334de04b50d32/write.go#L110

I need to extend this code to allow for single frame compression.

nhooyr:
I would suggest disabling compression till I get to this. Hopefully within the next 2 months.

andersroos:
> I would suggest disabling compression till I get to this. Hopefully within the next 2 months.

Thank you, will do.

andersroos:
> [@andersroos](https://github.com/andersroos), you've been waiting a while, but a fix was just merged in. No release yet, but it's there if you want to try it out.

Hey! Thank you!

Will test it for sure, but can't say if it will be before or after release yet.

kornes:
I ran into exact same issue with compressed messages, and can confirm #552 pr fixed it for me 👍

PR #552, opened 2025-12-17
fix: transmit in single frame when compression enabled

DanielleMaywood:
Closes https://github.com/coder/websocket/issues/435

### Problem

When compression was enabled, `Conn.Write` sent messages across many small frames due to the flate library's internal `bufferFlushSize` (240 bytes). Each flush triggered a `writeFrame` call, producing alternating ~236 and 4 byte frames. This broke clients like Unreal Engine that only process one frame per tick.

### Solution

`Conn.Write` now compresses the entire message into a buffer first, then transmits it as a single frame. Messages below `flateThreshold` bypass compression and are sent uncompressed in a single frame.

For `CompressionContextTakeover` mode, the flateWriter destination is restored after buffered compression to ensure subsequent `Writer()` streaming calls work correctly.

---

🤖 PR was written by Claude Code using Opus 4.5 then reviewed by Opus 4.5 and a human 👩‍🚀

### DIFF
--- a/write.go
+++ b/write.go
@@ -105,13 +105,12 @@ func (c *Conn) write(ctx context.Context, typ MessageType, p []byte) (int, error
 	if err != nil {
 		return 0, err
 	}
+	defer c.msgWriter.mu.unlock()
 
 	if !c.flate() || len(p) < c.flateThreshold {
-		defer c.msgWriter.mu.unlock()
 		return c.writeFrame(ctx, true, false, c.msgWriter.opcode, p)
 	}
 
-	defer c.msgWriter.mu.unlock()
 	return c.msgWriter.writeCompressedFrame(ctx, p)
 }
 


write.go
```diff
@@ -142,6 +143,46 @@ func (mw *msgWriter) putFlateWriter() {
...
+
+	// Buffer compressed output so we can write as a single frame instead
+	// of chunked frames.
+	origWriter := mw.trimWriter.w
+	mw.trimWriter.w = buf
+	defer func() {
+		mw.trimWriter.w = origWriter
+	}()
+
+	_, err := mw.flateWriter.Write(p)
```
mafredri: Bug: Using flateWriter without holding writeMu (Write implementation below gives a clue that this is necessary). If the connection closes right now, flateWriter may be nil.

write.go
```diff
@@ -110,13 +111,13 @@ func (c *Conn) write(ctx context.Context, typ MessageType, p []byte) (int, error
...
-	if err != nil {
-		return n, err
+	if len(p) < c.flateThreshold {
+		defer c.msgWriter.mu.unlock()
+		return c.writeFrame(ctx, true, false, c.msgWriter.opcode, p)
 	}
 
-	err = mw.Close()
-	return n, err
+	defer c.msgWriter.mu.unlock()
```
mafredri: Nit: This could be a single defer after reset.

compress_test.go
```diff
@@ -59,3 +63,232 @@ func BenchmarkFlateReader(b *testing.B) {
...
+	// 2. `Write` API
+	err := client.Write(ctx, MessageText, msg1)
+	assert.Success(t, err)
+
+	r := <-readCh
+	assert.Success(t, r.err)
+	assert.Equal(t, "msg1 type", MessageText, r.typ)
+	assert.Equal(t, "msg1 content", string(msg1), string(r.p))
+
+	// 2. `Writer` API
```
mafredri: Nit: Both comments say `2`, and they could be more descriptive if they're needed.

Should we also make this loop two times so we test all paths (write -> writer, writer -> write)?

compress_test.go
```diff
@@ -59,3 +63,232 @@ func BenchmarkFlateReader(b *testing.B) {
...
+		assert.Success(t, err)
+
+		withoutTakeoverSizes = append(withoutTakeoverSizes, h2.payloadLength)
+		assert.Success(t, <-writeDone2)
+	}
+
+	// With context takeover, the 2nd and 3rd messages should be smaller
+	// than without context takeover (dictionary helps compress repeated patterns).
+	// The first message will be similar size for both modes since there's no
+	// prior dictionary. But subsequent messages benefit from context takeover.
```
mafredri: 👍🏻

compress_test.go
```diff
@@ -59,3 +63,249 @@ func BenchmarkFlateReader(b *testing.B) {
...
+
+		_, err = io.CopyN(io.Discard, reader2, h2.payloadLength)
+		assert.Success(t, err)
+
+		withoutTakeoverSizes = append(withoutTakeoverSizes, h2.payloadLength)
+		assert.Success(t, <-writeDone2)
+	}
+
+	// With context takeover, the 2nd and 3rd messages should be smaller
+	// than without context takeover (dictionary helps compress repeated patterns).
```
mafredri:
```suggestion
	// With context takeover, the 2nd and 3rd messages should be smaller than
	// without context takeover (dictionary helps compress repeated patterns).
```
I know, I know..

PR #562, opened 2026-04-13
fix: do not wrap io.EOF in errd.Wrap

### LINKED ISSUE #561: Must not wrap io.EOF
From the Go docs:

```go
// EOF is the error returned by Read when no more input is available.
// (Read must return EOF itself, not an error wrapping EOF,
// because callers will test for EOF using ==.)
// Functions should return EOF only to signal a graceful end of input.
// If the EOF occurs unexpectedly in a structured data stream,
// the appropriate error is either [ErrUnexpectedEOF] or some other error
// giving more detail.
var EOF = errors.New("EOF")
```

however, `Conn.reader` (and probably others) wrap io.EOF:

```go
func (c *Conn) reader(ctx context.Context) (_ MessageType, _ io.Reader, err error) {
	defer errd.Wrap(&err, "failed to get reader")


// Wrap wraps err with fmt.Errorf if err is non nil.
// Intended for use with defer and a named error return.
// Inspired by https://github.com/golang/go/issues/32676.
func Wrap(err *error, f string, v ...any) {
	if *err != nil {
		*err = fmt.Errorf(f+": %w", append(v, *err)...)
	}
}
```

This leads to errors such as https://github.com/hashicorp/yamux/pull/167.

I'd propose special-casing `Wrap` to check for `io.EOF` and would be happy to open a PR.

Yanhu007:
Fixes #561

## Problem

The `errd.Wrap` helper wraps all non-nil errors using `fmt.Errorf("%w")`. When `io.EOF` is returned (e.g. from `Conn.reader`), it becomes `"failed to get reader: EOF"`. Per Go convention, `io.EOF` must never be wrapped:

> EOF is the error returned by Read when no more input is available.
> (Read must return EOF itself, not an error wrapping EOF,
> because callers will test for EOF using ==.)

This breaks callers like `io.Copy` and `bufio.Scanner` that test `err == io.EOF`.

## Fix

Skip wrapping in `errd.Wrap` when the error is exactly `io.EOF`:

```go
func Wrap(err *error, f string, v ...any) {
    if *err != nil && *err != io.EOF {
        *err = fmt.Errorf(f+": %w", append(v, *err)...)
    }
}
```

### DIFF
--- a/internal/errd/wrap.go
+++ b/internal/errd/wrap.go
@@ -2,13 +2,16 @@ package errd
 
 import (
 	"fmt"
+	"io"
 )
 
 // Wrap wraps err with fmt.Errorf if err is non nil.
 // Intended for use with defer and a named error return.
 // Inspired by https://github.com/golang/go/issues/32676.
+//
+// io.EOF is never wrapped because callers test for it with ==.
 func Wrap(err *error, f string, v ...any) {
-	if *err != nil {
+	if *err != nil && *err != io.EOF {
 		*err = fmt.Errorf(f+": %w", append(v, *err)...)
 	}
 }


internal/errd/wrap.go
```diff
@@ -2,13 +2,17 @@ package errd
...
 	"fmt"
+	"io"
 )
 
 // Wrap wraps err with fmt.Errorf if err is non nil.
 // Intended for use with defer and a named error return.
 // Inspired by https://github.com/golang/go/issues/32676.
+//
+// io.EOF is never wrapped because the Go convention requires
+// callers to test for EOF using ==, not errors.Is.
```
mafredri: This seems backwards. Go convention does not require callers to do this. The convention asks implementers to return io.EOF due to backwards compatibility and the stdlib implementation being frozen.

Callers should absolutely use `errors.Is`, it's the modern way.

mafredri (commented):
Hey, thanks for the PR!

I'm not convinced we should do this in `Wrap`, but rather handle this explicitly in `Read` methods. I can't think of a reason we should unwrap e.g. writes hitting `io.EOF`.

PR #566, opened 2026-06-15
conn: skip timeout callbacks for background contexts

mitchellh:
Skip per-operation timeout callback setup when the context cannot be canceled, such as `context.Background()`.

Timeout setup now reports whether it installed a callback, so callers only clear callbacks that exist. The benchmark also joins its writer goroutine to make repeated runs reliable.

Previously, reads and writes called `context.AfterFunc` for every operation, even when `ctx.Done()` was `nil`. This had a measurable cost.

Similar to #565, I know this is a micro-optimization, but its basic and understandable and it produced measurable results in some of my actual workloads that were cheap!

## Benchmarks

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| Header read, background context | 170.55 ns, 216 B, 6 allocs | 64.19 ns, 64 B, 2 allocs | -62.4% time, -70.4% bytes, -66.7% allocs |
| 512-byte payload read, background context | 305.5 ns, 216 B, 6 allocs | 190.1 ns, 64 B, 2 allocs | -37.8% time, -70.4% bytes, -66.7% allocs |

Each operation avoids 4 allocations and 152 bytes from the unnecessary `context.AfterFunc` registration.

Cancelable contexts are unchanged because they still require a cancellation callback. Accordingly, `BenchmarkConn/disabledCompress`, which uses a cancelable context, showed no statistically significant change.

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -168,24 +168,34 @@ func (c *Conn) close() error {
 	return err
 }
 
-func (c *Conn) setupWriteTimeout(ctx context.Context) {
+func (c *Conn) setupWriteTimeout(ctx context.Context) bool {
+	if ctx.Done() == nil {
+		return false
+	}
+
 	stop := context.AfterFunc(ctx, func() {
 		c.clearWriteTimeout()
 		c.close()
 	})
 	swapTimeoutStop(&c.writeTimeoutStop, &stop)
+	return true
 }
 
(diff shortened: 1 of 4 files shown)


kylecarbs (approved):

mitchellh:
Rebased.

kylecarbs (approved):

